PR #1, opened 2017-08-30
Change to Python

thewilkybarkid:
Only implements `/ping`.


.gitignore
```diff
@@ -1,2 +1,2 @@
-/vendor/
-/config.php
+/.cache/
+/venv/
```
giorgiosironi: no need for a `__pycache__` folder or `*.pyc` files? (varies between versions)

test/ping.py
```diff
@@ -0,0 +1,6 @@
+def test_it_pongs(test_client):
```
giorgiosironi: prefix to rename to `test_ping.py` (convention)

pytest.ini
```diff
@@ -0,0 +1,2 @@
+[pytest]
+python_files = test/*.py
```
giorgiosironi: there should be no need to have a `test/__init__.py` file, see https://docs.pytest.org/en/latest/goodpractices.html#tests-outside-application-code

profiles/__init__.py
```diff
@@ -0,0 +1 @@
+from .factory import create_app
```
giorgiosironi: this relative import makes you able to run `from profile import create_app` rather than `from profiles.factory import create_app`. So it's an additional layer of indirection, but not strictly needed

profiles/api/__init__.py
```diff
@@ -0,0 +1 @@
+from .ping import ping_bp
```
giorgiosironi: this allows `from profiles.api import ping_bp` rather than `from profiles.api.pin import ping_bp`, so in the same way can be useful to group all the objects that need to be exposed but it's only an aliasing operation

.gitignore
thewilkybarkid: Added `__pycache__`.

giorgiosironi:
Started building pull requests, which will fail until the formula is updated (work in progress)

PR #2, opened 2017-08-30
Adding pylint

giorgiosironi:
Be prepared to add exceptions like in https://github.com/elifesciences/elife-spectrum/blob/master/.pylintrc

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,2 @@
+[MESSAGE CONTROL]
+disable=missing-docstring
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,2 +1,2 @@
 [MESSAGE CONTROL]
-disable=missing-docstring
+disable=missing-docstring,redefined-outer-name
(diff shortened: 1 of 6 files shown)


giorgiosironi:
Not yet running into the build, local output:
```
************* Module profiles.api.ping
C:  3, 0: Invalid constant name "ping_bp" (invalid-name)
************* Module conftest
W: 11,16: Redefining name 'app' from outer scope (line 6) (redefined-outer-name)
************* Module test_ping
C:  4,11: Comparison should be response.status_code == 200 (misplaced-comparison-constant)
C:  5,11: Comparison should be response.data.decode('UTF-8') == 'pong' (misplaced-comparison-constant)
C:  6,11: Comparison should be response.headers.get('Content-Type') == 'text/plain; charset=UTF-8' (misplaced-comparison-constant)
```
1 is that with PEP 8 constants should be uppercase.
2 is a scope overriding a name from another module, unclear but not dangerous, we may override it
3 is the order of comparison, in Python is usually `actual == expected`

PR #6, opened 2017-09-01
Adds OAuth2

thewilkybarkid:
Depends on #2.

Testing with Google, but we'll be using ORCID.


test/test_oauth2.py
```diff
@@ -0,0 +1,234 @@
...
+                                data={'client_id': 'client_id', 'client_secret': 'client_secret',
+                                      'redirect_uri': 'http://www.example.com/client/redirect',
+                                      'grant_type': 'authorization_code'})
+
+    assert response.status_code == 400
+    assert response.headers.get('Content-Type') == 'application/problem+json'
+    assert json.loads(response.data.decode('UTF-8')) == {'title': 'Invalid code'}
+
+
+@responses.activate
```
thewilkybarkid: Is there anyway to turn on Responses for _all_ tests? I don't want to accidentally make any real requests.

profiles/factory.py
```diff
@@ -1,10 +1,33 @@
...
 from profiles.api.ping import PING_BP
+from werkzeug.exceptions import HTTPException, InternalServerError
 
 
-def create_app():
+def create_app(config):
     app = Flask(__name__)
+    app.TRAP_HTTP_EXCEPTIONS = True
+    app.config.update(dict(
+        config=config
```
thewilkybarkid: This doesn't feel like a great way of add configuration.

.pylintrc
```diff
@@ -1,2 +1,5 @@
 [MESSAGE CONTROL]
-disable=missing-docstring,redefined-outer-name
+disable=invalid-name,missing-docstring,redefined-outer-name
```
thewilkybarkid: Added `invalid-name` has it has a 30 character limit, and tests deserve expressive names. (Can it be disabled on `tests` only?)

config.yml.dist
```diff
@@ -0,0 +1,12 @@
+oauth2:
+  clients:
+    client-name:
+      id: id
+      secret: secret
+      redirect_uri: http://www.example.com/
```
thewilkybarkid: Normally multiple redirect locations are allowed, don't think that we've ever need more than 1 though?

config.yml.dist
```diff
@@ -0,0 +1,12 @@
...
+  server:
+    authorize_uri: http://www.example.com/auth
+    token_uri: http://www.example.com/token
+    client_id: client_id
+    client_secret: client_secret
+    scope: profile email
```
thewilkybarkid: Maybe shouldn't be in config?

profiles/factory.py
```diff
@@ -1,10 +1,36 @@
...
+    ))
 
+    app.register_blueprint(OAUTH2_BP, url_prefix='/oauth2')
     app.register_blueprint(PING_BP)
 
+    def http_error_handler(exception):
+        if not isinstance(exception, HTTPException):
+            exception = InternalServerError(getattr(exception, 'message', None))
+
+        if request.path.startswith('/oauth2/') and not request.path.startswith('/oauth2/token'):
```
thewilkybarkid: Is there a cleaner way to do this?

profiles/api/oauth2.py
```diff
@@ -0,0 +1,121 @@
...
+        raise BadRequest('Invalid client_id')
+
+    if request.args.get('redirect_uri') != client['redirect_uri']:
+        raise BadRequest('Invalid redirect_uri')
+
+    state = {
+        'redirect_uri': client['redirect_uri'],
+        'client_id': client['id'],
+        'original': request.args.get('state', '')
+    }
```
thewilkybarkid: Rather than having to pass all this, might be better to store locally and just include the key. Also gives CSRF protection for free (if random).

profiles/factory.py
thewilkybarkid: Some paths are meant to be interacted with a browser, others from applications. Better to split it on those lines rather than `Accept` (looking at the latter from a browser should still return JSON rather than HTML).

config.yml.dist
thewilkybarkid: Meant shouldn't be configurable, ie in code. (Profiles will be interacting with ORCID, not Journal etc.)

app.cfg.dist
```diff
@@ -0,0 +1,14 @@
+[oauth2]
+    [[clients]]
+        [[[client-name]]]
```
thewilkybarkid: This feels wrong, but not sure how to allow for nesting.

giorgiosironi: ah, `configobj`, it's the simplest solution 
Other people use convention:
`oauth2_clients_client_name: ...`

config.yml.dist
giorgiosironi: The `scope` itself could be a constant somewhere as it is likely the code would stop working if that config is changed?

profiles/factory.py
thewilkybarkid: In fact, it's more complicated than that. We need _two_ types of JSON errors: OAuth2 errors, and out standard API errors (RFC 7807).

thewilkybarkid: Make that _four_ types of error in total: HTML, redirect, and 2 JSON types. 😓

manage.py
```diff
@@ -1,8 +1,11 @@
+from configobj import ConfigObj
 from flask_script import Manager, Server, Shell
 
 from profiles.factory import create_app
 
-app = create_app()
+config = ConfigObj('app.cfg')
```
giorgiosironi: `.gitignore` should ignore this I think. The formula provides a file to be expanded: https://github.com/elifesciences/profiles-formula/blob/master/salt/profiles/config/srv-profiles-app.cfg

profiles/api/errors.py
```diff
@@ -0,0 +1,52 @@
+class OAuth2Error(Exception):
```
giorgiosironi: I don't think there's anything useful to extend as a base exception class in https://docs.python.org/3/library/exceptions.html but in other cases there could be

profiles/factory.py
```diff
@@ -1,10 +1,53 @@
...
+            body = remove_none_values({
+                'error': exception.error,
+                'error_description': exception.description,
+            })
+            return make_response(jsonify(body), exception.status_code)
+
+        if not isinstance(exception, HTTPException):
+            exception = InternalServerError(getattr(exception, 'message', None))
+
+        if request.path.startswith('/oauth2/authorize') or request.path.startswith('/oauth2/check'):
```
giorgiosironi: an option we have in the future is to send the user to one of the error URL provided in the configuration. We have that, but it's probably not easy to access the right one from here

manage.py
thewilkybarkid: Yep, it was already there.

profiles/api/errors.py
thewilkybarkid: I had wondered about `HTTPException`, but didn't feel right.

profiles/api/oauth2.py
```diff
@@ -0,0 +1,139 @@
...
+        raise ClientUnsupportedResourceType(client)
+
+    if request.args.get('scope'):
+        raise ClientInvalidScope(client)
+
+    state = remove_none_values({
+        'redirect_uri': client['redirect_uri'],
+        'client_id': client['id'],
+        'original': request.args.get('state')
+    })
```
thewilkybarkid: (Restoring still-valid comment.)

Rather than having to pass all this, might be better to store locally and just include the key. Also gives CSRF protection for free (if random).

thewilkybarkid: Maybe even store it in a session, rather than persisting it in a DB or key-value store?

giorgiosironi: Adding to the options: what about encrypting or signing it if it solves CSRF protection, without the persistence?

thewilkybarkid: Could sign it, but still it's getting quite a bit of data exposed in the URI. Let's create a ticket for it and do it later.

PR #7, opened 2017-09-05
Add type hints

thewilkybarkid:


profiles/api/ping.py
```diff
@@ -1,9 +1,13 @@
...
 @PING_BP.route('/ping')
-def ping():
-    return ('pong', {'Content-Type': 'text/plain; charset=UTF-8',
-                     'Cache-Control': 'must-revalidate, no-cache, no-store, private'})
+def ping() -> Response:
+    response = make_response('pong')
+    response.headers['Content-Type'] = 'text/plain; charset=UTF-8'
+    response.headers['Cache-Control'] = 'must-revalidate, no-cache, no-store, private'
+
+    return response
```
thewilkybarkid: A little more verbose, but less magic.

giorgiosironi: If only that were immutable

PR #8, opened 2017-09-06
Setting https on redirect_uri explicitly

giorgiosironi:
However, I wonder if it will use HTTPS when needed.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -21,7 +21,10 @@ def app() -> Flask:
                 'client_id': 'server_client_id',
                 'client_secret': 'server_client_secret',
             }
-        }
+        },
+        'service': {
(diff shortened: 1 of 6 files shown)


test/conftest.py
```diff
@@ -21,7 +21,10 @@ def app() -> Flask:
                 'client_id': 'server_client_id',
                 'client_secret': 'server_client_secret',
             }
-        }
+        },
+        'service': {
+            'default_host': 'localhost',
```
thewilkybarkid: Ok, but could be made the default anyway.

thewilkybarkid:
We should take a note to do this properly later (for dev/testing we may not want HTTPS), but good enough for now.

PR #10, opened 2017-09-07
ELPP-3141 Create eLife profile

thewilkybarkid:


.pylintrc
```diff
@@ -1,8 +1,9 @@
 [MESSAGE CONTROL]
-disable=missing-docstring,redefined-outer-name
+disable=missing-docstring,redefined-outer-name,too-few-public-methods
 
 [TYPECHECK]
-ignored-classes=responses
+ignored-classes=responses,scoped_session,SQLAlchemy
 
+good-names=db,id
```
giorgiosironi: :+1:

app.cfg.dist
```diff
@@ -1,3 +1,6 @@
+[db]
+uri = mysql+pymysql://username:password@hostname/profiles
```
giorgiosironi: any reason for MySQL vs Postgresql (used in many of our Python projects)?

migrations/env.py
```diff
@@ -0,0 +1,53 @@
+from alembic import context
+from flask import current_app
+from sqlalchemy import engine_from_config, pool
+from logging.config import fileConfig
+import logging
+
+config = context.config
+
+fileConfig(config.config_file_name)
+logger = logging.getLogger('alembic.env')
```
giorgiosironi: this is not a logger coming from the `alembic.env` module but used in our own code instead, I would go for `getLogger(__name__)`

migrations/script.py.mako
```diff
@@ -0,0 +1,16 @@
+from alembic import op
```
giorgiosironi: why does everyone invent its own templating language?
`Mako's syntax and API borrows from the best ideas of many others, including Django and Jinja2 templates, Cheetah, Myghty, and Genshi.`
hopefully we touch this template very little

migrations/versions/e1b6d52fd6ea_.py
```diff
@@ -0,0 +1,22 @@
...
+down_revision = None
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        'profile',
+        sa.Column('id', sa.String(length=8), nullable=False),
+        sa.Column('name', sa.String(length=128), nullable=False),
```
giorgiosironi: I assume if we update this name, we will keep the events in another table and this will be the latest `name`

migrations/versions/e1b6d52fd6ea_.py
```diff
@@ -0,0 +1,22 @@
...
+
+revision = 'e1b6d52fd6ea'
+down_revision = None
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        'profile',
```
giorgiosironi: convention to the singular?

migrations/versions/e1b6d52fd6ea_.py
giorgiosironi: I'll just read through http://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/ for pitfalls

migrations/versions/e1b6d52fd6ea_.py
```diff
@@ -0,0 +1,22 @@
...
+        sa.Column('id', sa.String(length=8), nullable=False),
+        sa.Column('name', sa.String(length=128), nullable=False),
+        sa.Column('orcid', sa.String(length=19), nullable=True),
+        sa.PrimaryKeyConstraint('id'),
+        sa.UniqueConstraint('orcid')
```
giorgiosironi: seems terse as a declarative language :+1:

profiles/api/oauth2.py
```diff
@@ -136,4 +138,14 @@ def token() -> Response:
     filtered_json_data = {your_key: json_data[your_key] for your_key in
                           ['access_token', 'expires_in', 'name', 'orcid', 'token_type']}
 
+    try:
+        profile = Profile.query.filter_by(orcid=filtered_json_data['orcid']).one()
+        profile.name = filtered_json_data['name']
+    except NoResultFound:
+        profile = Profile(filtered_json_data['name'], filtered_json_data['orcid'])
+
+        db.session.add(profile)
```
giorgiosironi: asymmetry: both `Profile.query` and `db.*` ultimately access persistence, but only one is named `db` :thinking:

profiles/api/oauth2.py
```diff
@@ -136,4 +138,14 @@ def token() -> Response:
     filtered_json_data = {your_key: json_data[your_key] for your_key in
                           ['access_token', 'expires_in', 'name', 'orcid', 'token_type']}
 
+    try:
+        profile = Profile.query.filter_by(orcid=filtered_json_data['orcid']).one()
```
giorgiosironi: does `.one()` mean that if there are 2 results, exception? (which shoudn't be possible due to constraints)

profiles/api/oauth2.py
```diff
@@ -136,4 +138,14 @@ def token() -> Response:
...
+        profile.name = filtered_json_data['name']
+    except NoResultFound:
+        profile = Profile(filtered_json_data['name'], filtered_json_data['orcid'])
```
giorgiosironi: we may need some sanitization, but the risk is mangling horribly someone's name. Welcome to eLife `dÃ©jÃn`!

profiles/api/oauth2.py
giorgiosironi: Repository pattern doesn't cut it without UnitOfWork too, we'll see later when there is more duplication

migrations/versions/e1b6d52fd6ea_.py
giorgiosironi: Is that 128 bytes or 128 Unicode chars? I guess the first

profiles/utilities.py
```diff
@@ -1,2 +1,14 @@
+import random
+import string
+
+
+def generate_id() -> str:
+    return generate_random_string(8, string.ascii_lowercase + string.digits)
```
giorgiosironi: uppercase excluded for readability?

giorgiosironi: 8 should be a constant somewhere application-wide, as it's used in the database schema definition too

profiles/api/oauth2.py
giorgiosironi: However, doesn't cover collisions? `id` is randomly generated, so it may already be in the database

test/conftest.py
```diff
@@ -1,12 +1,25 @@
...
 from flask import Flask
 from flask.testing import FlaskClient
+from flask_sqlalchemy import SQLAlchemy
+from sqlalchemy.orm import scoped_session
 from profiles.factory import create_app
+from profiles.models import db
 from pytest import fixture
 
+TEST_DATABASE_NAME = 'test.db'
+TEST_DATABASE_PATH = os.path.dirname(os.path.realpath(__file__)) + '/../build/' + TEST_DATABASE_NAME
```
giorgiosironi: ok, that folder should be writable by the testing framework running this

giorgiosironi: `ci` could run with `sqlite` and `end2end` with the real database

test/conftest.py
```diff
@@ -21,9 +34,53 @@ def app() -> Flask:
                 'client_id': 'server_client_id',
                 'client_secret': 'server_client_secret',
             }
-        }
+        },
     })
 
+    ctx = app.app_context()
+    ctx.push()
```
giorgiosironi: some comments on what global-state-restoring I suppose is performed here?

test/conftest.py
```diff
@@ -21,9 +34,53 @@ def app() -> Flask:
...
+@fixture(scope='function', autouse=True)
+def session(database: SQLAlchemy, request: FixtureRequest) -> scoped_session:
+    connection = database.engine.connect()
+    transaction = connection.begin()
+
+    options = dict(bind=connection, binds={})
+    session = database.create_scoped_session(options=options)
+
+    database.session = session
+
```
giorgiosironi: seems to be http://xunitpatterns.com/Transaction%20Rollback%20Teardown.html so I'd link it here

test/test_oauth2.py
```diff
@@ -219,6 +221,47 @@ def test_it_exchanges(test_client: FlaskClient):
...
+    db.session.add(original_profile)
+    db.session.commit()
+
+    test_client.post('/oauth2/token',
+                     data={'client_id': 'client_id', 'client_secret': 'client_secret',
+                           'redirect_uri': 'http://www.example.com/client/redirect',
+                           'grant_type': 'authorization_code', 'code': '1234'})
+
+    assert Profile.query.count() == 1
+    assert original_profile.name == 'Josiah Carberry'
```
giorgiosironi: doesn't hurt to update... for now

giorgiosironi:
In general, we probably need some domain modeling activity

profiles/api/oauth2.py
giorgiosironi: I will look for a dataset of names to check UTF-8 and possible conversions throughout the stack

profiles/api/oauth2.py
```diff
@@ -136,4 +138,14 @@ def token() -> Response:
     filtered_json_data = {your_key: json_data[your_key] for your_key in
                           ['access_token', 'expires_in', 'name', 'orcid', 'token_type']}
```
giorgiosironi: is `name` a placeholder or undocumented? https://members.orcid.org/api/get-oauthtoken doesn't list it, if it's the correct place to look?

migrations/env.py
thewilkybarkid: This directory is generated code.

profiles/api/oauth2.py
thewilkybarkid: Raises `MultipleResultsFound`.

profiles/api/oauth2.py
thewilkybarkid: Always useful. 👍

migrations/versions/e1b6d52fd6ea_.py
thewilkybarkid: I imagine so. Wonder what ORCID's length is. We might want to remove this for now, as they have first name, last name and published name.

profiles/api/oauth2.py
thewilkybarkid: True, should add a retry. (We'll need to override the ID generator in a test to do that.)

test/conftest.py
thewilkybarkid: Based on http://alexmic.net/flask-sqlalchemy-pytest/. Need to go through this part a bit better and work out what's happening.

profiles/api/oauth2.py
thewilkybarkid: Interesting, it's being returned but is undocumented. I think we'll have to use the real API at this point anyway, so can probably ignore it.

profiles/utilities.py
thewilkybarkid: Good idea. Not sure whether to pass in the generated ID, or move the `generate_random_string()` call into the object.

migrations/versions/e1b6d52fd6ea_.py
thewilkybarkid: Do we have a convention? Singular works until you need a `user` table.

migrations/script.py.mako
thewilkybarkid: Indeed.

app.cfg.dist
thewilkybarkid: I don't think it matters from the code perspective.

profiles/utilities.py
giorgiosironi: Here I was referring only to the `8` length constant, but identity generation is serious business (as what characterizes all Entities in Domain-Driven Design) and could use a Factory

migrations/versions/e1b6d52fd6ea_.py
giorgiosironi: singular good for me as `user` is not a domain term

app.cfg.dist
giorgiosironi: @lsh-0 any preference for an [dev]ops perspective?

migrations/versions/e1b6d52fd6ea_.py
thewilkybarkid: If we keep MySQL we should make sure that it's using `utf8mb4`.

app.cfg.dist
lsh-0: if it doesn't affect the code, then postgresql please.

migrations/script.py.mako
lsh-0: I was just thinking that looking at all that awful generated migration code. wtf.

migrations/versions/e1b6d52fd6ea_.py
lsh-0: > If we keep MySQL we should make sure that it's using utf8mb4
I wasn't going to mention any of the many reasons why we should avoid mysql in favour of postgresql, but it's ridiculous encoding problems would be near the top of my list.

migrations/env.py
thewilkybarkid: Wonder if we can actually exclude it...

thewilkybarkid: Doesn't look likely. 😞

lsh-0: no, you should be able to do a `logging.getLogger('alembic.env').setLevel(logging.ERROR)` (or something similar - we do it to quieten boto)

giorgiosironi: ok for silencing it if needed, on the other hand if it's code generated by the `alembic.env` module it makes sense it logs as such

profiles/api/oauth2.py
giorgiosironi: All popular datasets are US-centric, probably will generate random UTF-8 data... But need to check what ORCID returns first.

profiles/exceptions.py
```diff
@@ -1,6 +1,3 @@
-from profiles.models import Client
```
thewilkybarkid: Ran into a cyclical import. Changing `models.py` to not create `SQLAlchemy` looked like it would resolve it, but can't find out how to add the models (https://stackoverflow.com/questions/28789063/associate-external-class-model-with-flask-sqlalchemy).

giorgiosironi: Would a separate `repository.py` module help? Seems strange to have Entity and Repository interface and Repository implementation all in the same place. `exceptions` and `repository` could depend on `models`.
Moreover, `Client` doesn't look like part of the domain but only a configuration class (but don't know what depends on it)

thewilkybarkid: Resolved

profiles/repositories.py
```diff
@@ -0,0 +1,61 @@
...
+        except NoResultFound as exception:
+            raise ProfileNotFound('Profile with the ORCID {} not found'.format(orcid)) \
+                from exception
+
+    def next_id(self) -> str:
+        while True:
+            profile_id = generate_random_string(ID_LENGTH, string.ascii_lowercase + string.digits)
+            try:
+                self.get(profile_id)
+            except ProfileNotFound:
```
giorgiosironi: uhm http://wiki.c2.com/?DontUseExceptionsForFlowControl

migrations/versions/e1b6d52fd6ea_.py
thewilkybarkid: In Postgres (and MySQL 5) it's chars not bytes.

Guess https://github.com/ORCID/ORCID-Source/blob/a78c2ad843943ab0a4060a536118d4b8f4e7ab1e/orcid-persistence/src/main/resources/schema.sql#L476-L477 means that the ORCID name could be 301 chars.

giorgiosironi: curiously, these limits are hardcoded in the schema
https://github.com/ORCID/ORCID-Source/blob/master/orcid-model/src/main/resources/record_2.0/personal-details-2.0.xsd#L110-L133
along with the interesting
https://github.com/ORCID/ORCID-Source/blob/master/orcid-model/src/main/resources/common_2.0/common-2.0.xsd#L1440-L1457

thewilkybarkid: Huh. Just seen that the DB length looks to have been removed - https://github.com/ORCID/ORCID-Source/blob/5f5508183ef971d2d7529af7d6a84554aa860749/orcid-persistence/src/main/resources/db/updates/varchar_to_text.xml#L152-L156

giorgiosironi: this is not a field that should be indexed, so using an arbitrary `text` shouldn't have a big impact?

thewilkybarkid: We might want searching by name in the future (I'm envisaging someone using Journal CMS), but quickly reading http://rachbelaid.com/postgres-full-text-search-is-good-enough/ doesn't look like it'd be a problem. So let's change it.

thewilkybarkid:
@giorgiosironi Any final thoughts?

profiles/factory.py
```diff
@@ -1,17 +1,28 @@
...
     app.TRAP_HTTP_EXCEPTIONS = True
     app.config.from_object(config)
 
-    app.register_blueprint(oauth2.create_blueprint(config.orcid, clients), url_prefix='/oauth2')
+    db.app = app
+    db.init_app(app)
+
+    Migrate(app, db)
+
+    profiles = SQLAlchemyProfiles(db)
```
giorgiosironi: good injection of the Repository

profiles/models.py
```diff
@@ -1,37 +1,19 @@
...
-    elements = []
 
-    def __init__(self, iterable: Iterable[Client]) -> None:
-        for value in iterable:
-            if value not in self.elements:
-                self.elements.append(value)
+class Profile(db.Model):
+    id = db.Column(db.String(ID_LENGTH), primary_key=True)
+    name = db.Column(db.Text(), nullable=False)
+    orcid = db.Column(db.String(19), unique=True)
```
giorgiosironi: from yesterday, we may want to introduce immediately a `state` field

profiles/repositories.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def get_by_orcid(self, orcid: str) -> Profile:
+        try:
+            return self.db.session.query(Profile).filter_by(orcid=orcid).one()
+        except NoResultFound as exception:
+            raise ProfileNotFound('Profile with the ORCID {} not found'.format(orcid)) \
+                from exception
+
+    def next_id(self) -> str:
+        while True:
```
giorgiosironi: perhaps limiting the number of iterations and throwing an exception in case the limit is hit will prevent bugs in the filter from escalating to an infinite loop

profiles/api/oauth2.py
```diff
@@ -122,9 +125,16 @@ def _token() -> Response:
...
-                              ['access_token', 'expires_in', 'name', 'orcid', 'token_type']}
+        json_data = {key: json_data[key] for key in
+                     ['access_token', 'expires_in', 'name', 'orcid', 'token_type']}
 
-        return make_response(jsonify(filtered_json_data), response.status_code)
+        try:
+            profile = profiles.get_by_orcid(json_data['orcid'])
+            profile.name = json_data['name']
+        except ProfileNotFound:
+            profile = Profile(profiles.next_id(), json_data['name'], json_data['orcid'])
```
giorgiosironi: is the session automatically committed (assuming there is an open transaction) at the end of the request? when in particular?

profiles/models.py
thewilkybarkid: I did think about state and history but wanted to keep this PR as small as needed. (Since it doesn't cover deleting etc.)

giorgiosironi:
Main blocker is that we'll have to provision a local Postgresql database in `ci` and RDS instances for `end2end` and `prod`. Time for some knowledge sharing about that...

profiles/repositories.py
thewilkybarkid: Probably a good idea. Timeouts should normally kick in, but no need to get that far.

profiles/api/oauth2.py
thewilkybarkid: End of the request (`SQLALCHEMY_COMMIT_ON_TEARDOWN`). It's a bit of a controversial technique, but transactions are hard with repositories. (Never found a good solution.)

It's probably acceptable here, since it does actually do that much.

giorgiosironi: We'll go for an explicit Unit Of Work if needed

giorgiosironi:
Idiomatic Python 3 checking would be interesting (e.g. local issues that Pylint doesn't find but humans may)
I think we need to get down the architecture on paper (more "on bytes") in the form of [ADR](http://thinkrelevance.com/blog/2011/11/15/documenting-architecture-decisions) but it's early to do that, as there is not even a database in the formula yet.

PR #12, opened 2017-09-11
Fix non-HTTPException error handling

thewilkybarkid:
Not sure how PyCharm is running pytest differently, but it was showing some non HTTPExceptions not being handled (but still as  passes; `project_tests.sh` just showed passes).

### DIFF
--- a/profiles/factory.py
+++ b/profiles/factory.py
@@ -4,7 +4,7 @@ from flask import Flask, jsonify, make_response, redirect, request
 from profiles.api.errors import OAuth2Error, ClientError
 from profiles.api.oauth2 import OAUTH2_BP
 from profiles.api.ping import PING_BP
-from profiles.utilities import remove_none_values
+from profiles.utilities import remove_none_values, chain_exception
 from werkzeug.exceptions import HTTPException, InternalServerError
 from werkzeug.wrappers import Response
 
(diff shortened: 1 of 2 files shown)


profiles/utilities.py
```diff
@@ -1,2 +1,11 @@
+from typing import Type
+
+
+def chain_exception(exception: Type[Exception], previous: Exception, message: str = None):
+    exception = exception(message or str(previous))
+    exception.__cause__ = previous
```
thewilkybarkid: See also #11. `from` only works with `raise`.

giorgiosironi: I take this in the sense that it cannot be used to create an exception object outside of the `raise` construct, where it does as syntactic sugar what is being done here

thewilkybarkid: Yep. This is a little redundant now as we don't actually do anything with the previous exception, but when we add logging etc it will be useful.

profiles/factory.py
```diff
@@ -4,7 +4,7 @@
 from profiles.api.errors import OAuth2Error, ClientError
 from profiles.api.oauth2 import OAUTH2_BP
 from profiles.api.ping import PING_BP
-from profiles.utilities import remove_none_values, chain_exception
+from profiles.utilities import chain_exception, remove_none_values
```
thewilkybarkid: Any way to enforce this in Pyline @giorgiosironi?

giorgiosironi: This may be considered style rather than static analysis, as the semantics of the two are the same; should check out https://github.com/elifesciences/builder/blob/master/.scrub.sh#L11

PR #13, opened 2017-09-11
Check import orders

thewilkybarkid:


.pylintrc
```diff
@@ -1,8 +1,8 @@
 [MESSAGE CONTROL]
-disable=missing-docstring,redefined-outer-name
+disable = missing-docstring,redefined-outer-name
```
thewilkybarkid: Just matching INI styles elsewhere.

profiles/api/oauth2.py
```diff
@@ -1,13 +1,15 @@
...
 from urllib.parse import urlencode
+
+from flask import Blueprint, current_app, json, jsonify, make_response, redirect, request, url_for
 import requests
-from flask import Blueprint, current_app, make_response, redirect, request, jsonify, url_for, json
-from profiles.api.errors import InvalidClient, InvalidRequest, UnsupportedGrantType, InvalidGrant, \
-    ClientInvalidRequest, ClientUnsupportedResourceType, ClientInvalidScope
-from profiles.utilities import remove_none_values
 from werkzeug.exceptions import BadRequest
 from werkzeug.wrappers import Response
```
giorgiosironi: agreed, application code imported after libraries

giorgiosironi: (and the order of imports in the same line ofc)

thewilkybarkid:
This doesn't mandate a line between system/libraries and libraries/application imports. Possible?

PR #14, opened 2017-09-11
Improve configuration and code style

thewilkybarkid:


lsh-0:
can you switch profiles over to using python 3.5 please? 3.6 isn't part of 16.04, these python ppas haven proven themselves flaky and we've had issues with newrelic/uwsgi/django + python3.6

lsh-0:
:+1: 

cheers

profiles/config.py
```diff
@@ -0,0 +1,40 @@
...
+        environment = 'dev'
+
+    if environment == 'dev':
+        return DevConfig(config['orcid'])
+    elif environment == 'ci':
+        return CiConfig(config['orcid'])
+    if environment == 'prod':
+        return ProdConfig(config['orcid'])
+    if environment == 'end2end':
+        return End2EndConfig(config['orcid'])
```
thewilkybarkid: This doesn't seem great.

profiles/api/ping.py
```diff
@@ -1,13 +1,16 @@
...
+def create_blueprint() -> Blueprint:
+    blueprint = Blueprint('ping', __name__)
 
-@PING_BP.route('/ping')
-def ping() -> Response:
-    response = make_response('pong')
-    response.headers['Content-Type'] = 'text/plain; charset=UTF-8'
-    response.headers['Cache-Control'] = 'must-revalidate, no-cache, no-store, private'
+    @blueprint.route('/ping')
+    def _ping() -> Response:
```
thewilkybarkid: Run into https://github.com/pallets/flask/issues/2434. Don't want to disable the check, but means all routes with have to begin with `_`.

giorgiosironi: :+1: as not ideal but not wrong: immediately clear that this identifier is not available at the module level but only inside this scope

thewilkybarkid: Does get used inside `url_for()` though.

test/conftest.py
```diff
@@ -2,28 +2,26 @@
...
                 'token_uri': 'http://www.example.com/server/token',
                 'client_id': 'server_client_id',
                 'client_secret': 'server_client_secret',
-            }
-        }
-    })
+            },
+        ),
+        clients=Clients([
+            Client('client', 'client_id', 'client_secret', 'http://www.example.com/client/redirect'),
```
giorgiosironi: could probably use named arguments and multiple lines to be
- both shorter
- more clear

clients.json.dist
```diff
@@ -0,0 +1,7 @@
+{
+    "client-name": {
```
giorgiosironi: JSON is easy to parse, in general I would reserve it to files that are written by programs rather than configuration as it doesn't support comments, the last element `,`, etc... I realize every format has its own defects

manage.py
```diff
@@ -1,20 +1,28 @@
...
+from profiles.config import create_config
 from profiles.factory import create_app
+from profiles.models import Client, Clients
+
+config = configparser.ConfigParser()
+config.read('app.cfg')
 
-config = ConfigObj('app.cfg')
+clients_json = json.load(open('clients.json'))
+clients = Clients([Client(name, **clients_json[name]) for name in clients_json])
```
giorgiosironi: :+1: for strong typing

manage.py
```diff
@@ -1,20 +1,28 @@
...
 
-app = create_app(config)
+app = create_app(create_config(config), clients)
```
giorgiosironi: weird to pass `config` to `create_config`?

profiles/config.py
```diff
@@ -0,0 +1,38 @@
...
+from typing import Dict
+
+
+class Config(object):
+    def __init__(self, orcid: Dict[str, str]) -> None:
+        self.orcid = orcid
+
+
+class DevConfig(Config):
+    DEBUG = True
```
giorgiosironi: shouldn't there be a `False` default for these constants in the base `Config`?

profiles/config.py
giorgiosironi: build a `Dict[str, type]` mapping name to class?

profiles/factory.py
```diff
@@ -1,55 +1,25 @@
...
-
-        body = {
-            'title': getattr(exception, 'description', exception.name),
-        }
-        response = make_response(jsonify(body), exception.code)
-        response.headers['Content-Type'] = 'application/problem+json'
+    app.config.from_object(config)
 
-        return response
+    app.register_blueprint(oauth2.create_blueprint(config.orcid, clients), url_prefix='/oauth2')
```
giorgiosironi: I love the smell of Dependency Injection in the morning

test/test_oauth2.py
```diff
@@ -80,44 +80,44 @@ def test_it_redirects_with_the_original_state_when_authorizing(test_client: Flas
...
-def test_it_requires_a_code_when_checking(test_client: FlaskClient):
+def test_it_requires_a_code_when_checking(test_client: FlaskClient) -> None:
     response = test_client.get('/oauth2/check')
 
     assert response.status_code == 400
     assert 'Invalid code' in response.data.decode('UTF-8')
 
 
-def test_it_requires_a_state_when_checking(test_client: FlaskClient):
+def test_it_requires_a_state_when_checking(test_client: FlaskClient) -> None:
```
giorgiosironi: is that needed for every method? :java:

giorgiosironi: (considering no code calls test methods explicitly)

lsh-0: this is the same kind of thinking that leads to OO and getters and setters and code generation and mutable state hiding in deeply nested class hierarchies and .... argh! I have to ignore this code, it's making me angry.

profiles/config.py
thewilkybarkid: Might be a bit clearer. Flask has a default of `False` though.

clients.json.dist
thewilkybarkid: Yeah, I normally use YAML. Will take a look.

manage.py
thewilkybarkid: Yeah, `create_app_config(config_file)`?

profiles/config.py
thewilkybarkid: What happens when one environment needs different input? Pass in `**config` instead?

test/test_oauth2.py
giorgiosironi: OO has been the opposite of getters and setters for the last 14 years https://www.javaworld.com/article/2073723/core-java/why-getter-and-setter-methods-are-evil.html :)

PR #17, opened 2017-09-14
Fix nondeterminism in tests

thewilkybarkid:
Since order doesn't matter, an alternative is to compare normalised values.


thewilkybarkid:
Looks like Python 3.6 makes dictionaries ordered, but not by contract. So this only became visible following the downgrade to 3.5.

PR #19, opened 2017-09-15
Enable New Relic

thewilkybarkid:


giorgiosironi:
Due to Python 3 + uwsgi + New Relic weird combinations, we started using https://github.com/elifesciences/lax/blob/develop/install.sh#L25 in lax to avoid installing some compiled extensions that caused crashes. Good to go for now, we'll see if we need that here.

PR #20, opened 2017-09-15
ELPP-2775 Store ORCID access token

thewilkybarkid:


profiles/models.py
```diff
@@ -1,10 +1,23 @@
...
 from flask_sqlalchemy import SQLAlchemy
 
 db = SQLAlchemy()
 
 ID_LENGTH = 8
 
 
+class OrcidToken(db.Model):
+    orcid = db.Column(db.String(19), primary_key=True)
+    access_token = db.Column(db.String(255), unique=True)
```
giorgiosironi: why nullable on `access_token`?

profiles/repositories.py
```diff
@@ -29,6 +39,21 @@ def next_id(self) -> str:
         raise NotImplementedError
 
 
+class SQLAlchemyOrcidTokens(OrcidTokens):
+    def __init__(self, db: SQLAlchemy) -> None:
+        self.db = db
```
giorgiosironi: why not injecting only `self.db.session`? It has a too short lifetime or will `self.db` soon be needed?

profiles/utilities.py
```diff
@@ -1,5 +1,10 @@
 import random
 import string
+from datetime import datetime, timedelta
+
+
+def expires_at(expires_in: int) -> datetime:
```
giorgiosironi: don't understand why the spec put in a `expires_in` which makes the expiration time inaccurate while they could have provided a precise `expires_at`

profiles/repositories.py
thewilkybarkid: Just following code seen elsewhere. Reading the docs it looks safe to inject `db.sessions`.

profiles/models.py
thewilkybarkid: Good spot.

PR #23, opened 2017-09-20
Include ID in token response

thewilkybarkid:
Clients should use the eLife ID rather than ORCID (though I've left in the original details).

Related: other OAuth2 implementers seem to require you to make a second request to find out such details, ORCID don't (and include the `orcid` property).

### DIFF
--- a/test/test_oauth2.py
+++ b/test/test_oauth2.py
@@ -1,4 +1,5 @@
 from json import dumps, loads
+from unittest.mock import MagicMock, patch
 from urllib.parse import urlencode
 
 from flask.testing import FlaskClient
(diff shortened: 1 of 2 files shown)


test/test_oauth2.py
```diff
@@ -221,7 +225,7 @@ def test_it_exchanges(test_client: FlaskClient) -> None:
     assert loads(response.data.decode('UTF-8')) == {'access_token': '1/fFAGRNJru1FTz70BzhT3Zg',
                                                     'expires_in': 3920, 'token_type': 'Bearer',
                                                     'orcid': '0000-0002-1825-0097',
-                                                    'name': 'Josiah Carberry'}
+                                                    'name': 'Josiah Carberry', 'id': '1a2b3c4e'}
```
giorgiosironi: any need for a more specific name than `id`?

PR #24, opened 2017-09-20
Allow additional, unused configuration

giorgiosironi:
For forward compatibility, allow the formula to provide more configuration that needed (e.g. a `logging` key). This configuration is ignored.


giorgiosironi:
As soon as we have logging :snake:

PR #26, opened 2017-09-21
ELPP-2775 Fetch email addresses from ORCID

thewilkybarkid:
Still very much under development.


profiles/orcid.py
```diff
@@ -0,0 +1,25 @@
+import json
+
+import requests
+from requests import Response
+
+API_VERSION = 'v2.0'
+
+
+class OrcidClient(object):
```
thewilkybarkid: Doesn't look to be a usable Python library for the 2.0 API. https://github.com/ORCID/python-orcid looks unstable.

test/test_orcid.py
```diff
@@ -0,0 +1,28 @@
...
+import responses
+from requests import HTTPError
+
+from profiles.orcid import OrcidClient
+
+
+@responses.activate
+def test_it_gets_a_record():
+    responses.add(responses.GET, 'http://www.example.com/api/v2.0/0000-0002-1825-0097/record',
+                  status=200, json={'foo': 'bar'})
```
thewilkybarkid: Doesn't look to be a way to check the rights headers were on the request (`Accept` and `Authorization`).

thewilkybarkid:
Pylint currently complaining with a bunch of `no-name-in-module`, not sure why.

profiles/orcid.py
giorgiosironi: Agreed, you don't want to be cut-off from a functionality because a library is not up-to-date.

profiles/api/oauth2.py
```diff
@@ -147,6 +152,12 @@ def _token() -> Response:
                                      expires_at(json_data['expires_in']))
             orcid_tokens.add(orcid_token)
 
+        try:
+            orcid_record = orcid_client.get_record(profile.orcid, orcid_token.access_token)
+            profile.update_from_orcid_record(orcid_record)
+        except RequestException as exception:
+            LOGGER.exception(exception)
```
giorgiosironi: `logger.exception` takes a message, not the exception (which already has due to global state)

thewilkybarkid: `errors.py` looks wrong too?

profiles/models.py
```diff
@@ -22,11 +23,74 @@ class Profile(db.Model):
...
+    email_addresses = db.relationship('EmailAddress', order_by='EmailAddress.position',
+                                      collection_class=ordering_list('position'),
+                                      cascade='all, delete-orphan', back_populates='profile')
 
     def __init__(self, profile_id: str, name: str, orcid: str = None) -> None:
         self.id = profile_id
         self.name = name
         self.orcid = orcid
 
+    def update_from_orcid_record(self, orcid_record: dict):
```
giorgiosironi: not sure arbitrarily traversing the orcid data structures belong in the domain layer, could be extracted as a OrcidRecordUpdate Command class

profiles/models.py
```diff
@@ -22,11 +23,74 @@ class Profile(db.Model):
...
+                    break
+            if not found:
+                self.remove_email_address(email.email)
+
+        for orcid_email in orcid_record['person']['emails']['email'] or {}:
+            if orcid_email['verified']:
+                self.add_email_address(orcid_email['email'], orcid_email['primary'],
+                                       orcid_email['visibility'] != 'PUBLIC')
+
+    def add_email_address(self, email: str, primary: bool = False,
```
giorgiosironi: this synchronization primitives definitely belong here instead

profiles/orcid.py
```diff
@@ -0,0 +1,25 @@
...
+        response = self._get_request('{}/record'.format(orcid), access_token)
+
+        return json.loads(response.text)
+
+    def _get_request(self, path, access_token) -> Response:
+        uri = '{}/{}/{}'.format(self.api_uri, API_VERSION, path)
+        headers = {'Accept': 'application/orcid+json', 'Authorization': 'Bearer ' + access_token}
+
+        response = requests.get(uri, headers=headers)
+        response.raise_for_status()
```
giorgiosironi: this would be a good place for explicit logging

test/conftest.py
```diff
@@ -16,12 +17,15 @@
 TEST_DATABASE_PATH = os.path.dirname(os.path.realpath(__file__)) + '/../build/' + TEST_DATABASE_NAME
 TEST_DATABASE_URI = 'sqlite:///' + TEST_DATABASE_PATH
 
+logging.disable(logging.CRITICAL)
```
giorgiosironi: unfortunately idiomatic

test/test_oauth2.py
```diff
@@ -234,6 +234,40 @@ def test_it_creates_a_profile_when_exchanging(test_client: FlaskClient) -> None:
                   json={'access_token': '1/fFAGRNJru1FTz70BzhT3Zg', 'expires_in': 3920,
                         'foo': 'bar', 'token_type': 'Bearer', 'orcid': '0000-0002-1825-0097',
                         'name': 'Josiah Carberry'})
+    responses.add(responses.GET, 'http://www.example.com/api/v2.0/0000-0002-1825-0097/record',
```
giorgiosironi: we are at the limit of what can be tested at the API level, a single email address should be fine; but more complex cases can be tested at the domain or (if the email update command is extracted) immediately over that, without involving the outer API and the token exchange

thewilkybarkid: Yeah, extracting `update_from_orcid_record()` would make such testing easier. And adding unit tests for `Profile`.

test/test_orcid.py
giorgiosironi: https://pypi.python.org/pypi/requests-mock seems to allow matching on headers
Perhaps it's easier to keep it simple and `@patch('requests.get')`

test/test_oauth2.py
thewilkybarkid: Begun in https://github.com/elifesciences/profiles/pull/27.

profiles/api/oauth2.py
giorgiosironi: yes, it is

profiles/orcid.py
thewilkybarkid: All responses as `DEBUG`?

giorgiosironi: since for now it's only `GET`, `DEBUG`; if we had `POST` or similar, would go for an higher level

thewilkybarkid: `requests` logs HTTP requests/responses anyway, maybe better to log messages like 'Failed to load record'.

giorgiosironi: have yet to see a comprehensive log from `requests` and underlying library, let's wait

thewilkybarkid: Added https://github.com/elifesciences/profiles/pull/26/commits/01f692e4cbefcad5d33b0b36ea19afda7c8a2c8f.

profiles/utilities.py
```diff
@@ -24,32 +22,3 @@ def guess_index_name(name: str) -> str:
 
 def remove_none_values(items: dict) -> dict:
     return dict(filter(lambda item: item[1] is not None, items.items()))
-
-
-def update_profile_from_orcid_record(profile: Profile, orcid_record: dict) -> None:
```
thewilkybarkid: Moved to `commands.py` to avoid a cyclical import.

giorgiosironi:
ORCID won't give us access to production data until they can demo this, so we are going to switch to their sandbox, deploy and implement a few other things.

giorgiosironi:
Depends on https://github.com/elifesciences/builder-private/pull/57 now.

PR #27, opened 2017-09-21
Add unit tests

thewilkybarkid:


profiles/api/errors.py
```diff
@@ -40,10 +41,10 @@ def http_error_handler(exception: HTTPException) -> Response:
 def oauth2_error_handler(exception: OAuth2Error) -> Response:
     LOGGER.exception(exception)
 
-    body = remove_none_values({
-        'error': exception.error,
-        'error_description': exception.description,
-    })
+    body = remove_none_values(OrderedDict([
```
thewilkybarkid: Turns out #17 missed one.

profiles/models.py
```diff
@@ -1,22 +1,23 @@
...
+from profiles.database import db, UTCDateTime
 
 ID_LENGTH = 8
 
 
 class OrcidToken(db.Model):
     orcid = db.Column(db.String(19), primary_key=True)
     access_token = db.Column(db.String(255), unique=True, nullable=False)
-    expires_at = db.Column(db.DateTime(), nullable=False)
+    expires_at = db.Column(UTCDateTime, nullable=False)
```
thewilkybarkid: SQLAlchemy only seems to use naive datetimes, whereas we really want to make sure we talk UTC everywhere.

profiles/database.py
```diff
@@ -0,0 +1,27 @@
...
+
+class UTCDateTime(types.TypeDecorator):
+    impl = types.DateTime
+
+    def python_type(self) -> Type:
+        return datetime
+
+    def process_bind_param(self, value, dialect) -> Optional[datetime]:
+        if value is not None:
+            return value.astimezone(timezone.utc)
```
giorgiosironi: if I understand correctly, this decorator will transform the timezone of a `datetime` object when it is stored into the database, and then assume UTC when it takes it out.
However, this could lead to different behavior where a `OrcidToken` object is assigned a `datetime` object with the wrong timezone, and is used before saving it in the same process? In that case it would show the original, wrong timezone?
This normalization could be done in the domain object instead, whenever the field is assigned (as it shouldn't be assigned without indirection from outside the object).

profiles/models.py
```diff
@@ -1,22 +1,23 @@
...
 
     def __init__(self, orcid: str, access_token: str, expires_at: datetime) -> None:
         self.orcid = orcid
         self.access_token = access_token
         self.expires_at = expires_at
```
giorgiosironi: normalization of timezone could take place here

test/models/test_orcid_token.py
```diff
@@ -0,0 +1,26 @@
+from profiles.models import OrcidToken
+from profiles.utilities import expires_at
+
+
+def test_it_can_be_printed():
+    orcid_token = OrcidToken('0000-0002-1825-0097', '1/fFAGRNJru1FTz70BzhT3Zg', expires_at(1234))
+
+    assert repr(orcid_token) == "<OrcidToken '0000-0002-1825-0097'>"
```
giorgiosironi: I assume it's the safest option to avoid printing the token value implicitly, in case it gets logged or dumped somewhere public

test/repositories/sql_alchemy/test_profiles.py
```diff
@@ -0,0 +1,71 @@
...
+    profiles.add(Profile('11111111', 'name'))
+
+    generate_random_string.side_effect = ['11111111', '11111112']
+
+    assert profiles.next_id() == '11111112'
+
+
+@patch('profiles.repositories.generate_random_string')
+def test_it_limits_retrying_when_generating_the_next_profile_id(generate_random_string: MagicMock):
+    profiles = SQLAlchemyProfiles(db)
```
giorgiosironi: `generate_random_string` can easily be injected into `SQLAlchemyProfiles(db, generate)` to avoid the `@patch`

giorgiosironi: `generate = Mock()`

test/models/test_orcid_token.py
thewilkybarkid: That's the ORCID rather than the access token.

thewilkybarkid: (Could add 'for'.)

giorgiosironi: yes, that's why this way is safe :+1:

giorgiosironi:
Working with a branch through composer is painful, especially since this needs recompilation through `node compile.js`, maybe we should merge this by now with a `not-implemented` annotation?

profiles/database.py
thewilkybarkid: That's true, will change on initialisation too.

profiles/clients.py
```diff
@@ -14,10 +14,9 @@ def __repr__(self) -> str:
 
 
 class Clients(collections.Set):
-    elements = []
```
thewilkybarkid: Seems that the list is static, and so every instance appends to it?

profiles/database.py
thewilkybarkid: This type is needed, unless we have a getter that converts it to UTC.

giorgiosironi: So the type should only convert on reading but does not need to intervene on writing?

thewilkybarkid: Correct. Think we can remove this method.

thewilkybarkid: Hmm, the linting gods demand it. `NotImplementedError` works.

thewilkybarkid: Or disable `abstract-method`. I guess tests would fail if things aren't implemented correctly.

profiles/clients.py
giorgiosironi: that is a class-level variable, yes

PR #28, opened 2017-09-22
Use requests-mock rather than responses

thewilkybarkid:
Switches `responses` to `requests-mock` as it allows us match on the request body, headers etc.

Also adds `pytest-socket` to block _all_ HTTP requests that haven't been mocked, rather than just cases where the mocker is enabled.

### DIFF
--- a/test/test_oauth2.py
+++ b/test/test_oauth2.py
@@ -4,7 +4,8 @@ from urllib.parse import urlencode
 
 from flask.testing import FlaskClient
 from freezegun import freeze_time
-import responses
+from requests import Request
+import requests_mock
 
 from profiles.models import OrcidToken, Profile, db
 from profiles.utilities import expires_at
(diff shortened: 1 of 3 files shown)


test/test_oauth2.py
```diff
@@ -206,19 +207,29 @@ def test_it_requires_code_when_exchanging(test_client: FlaskClient) -> None:
...
-    responses.add(responses.POST, 'http://www.example.com/server/token', status=200,
-                  json={'access_token': '1/fFAGRNJru1FTz70BzhT3Zg', 'expires_in': 3920,
-                        'foo': 'bar', 'token_type': 'Bearer', 'orcid': '0000-0002-1825-0097',
-                        'name': 'Josiah Carberry'})
-
-    response = test_client.post('/oauth2/token',
-                                data={'client_id': 'client_id', 'client_secret': 'client_secret',
-                                      'redirect_uri': 'http://www.example.com/client/redirect',
-                                      'grant_type': 'authorization_code', 'code': '1234'})
+    with requests_mock.Mocker() as mocker:
```
thewilkybarkid: Can't use the decorator due to https://github.com/pytest-dev/pytest/issues/2749.

giorgiosironi: at least it shows the phase in which the mock is called

jamielennox:
Hey, by default requests-mock will raise a `requests_mock.NoMockAddress` exception for anything that hasn't been explicitly mocked out so depending on how you structure your tests you shouldn't need to do the pytest-socket.

thewilkybarkid:
Hi @jamielennox, it was just for cases where requests-mock wasn't activated (so there's no chance of a real HTTP request being made anywhere). You're right that there's probably a better structure that I'm not aware of though, do you know of any good examples?

jamielennox:
So I mostly use testtools for my tests and what i do is just start the mocker in the setUp() that runs before every test in a base class, even if i'm not going to use requests-mock in that test at all. It's not going to hurt anything to be running but unused. 

Unfortunately though i'm quite new to pytest and don't know how that could be replicated. Because the mocker object is recording the requests that are sent to it you need to start/stop it per test, not globally.

giorgiosironi:
Perhaps https://docs.pytest.org/en/latest/writing_plugins.html#_pytest.hookspec.pytest_runtest_setup could help.

PR #29, opened 2017-09-25
Store index name

thewilkybarkid:
Our API has a concept of a sortable name, as well as the display ('preferred') name. ~For flexibility we should store the name parts (rather than the structure that will be served through the API).~


thewilkybarkid:
Build fails as the DB migration conflicts with #26. We should remove all tables before each build.

giorgiosironi:
Now builds, some non-deterministic test failing on the query string parameters.

thewilkybarkid:
Yep, pulling in https://github.com/elifesciences/profiles/pull/26/commits/8bdf85f3301d0d405410969036426ec6e989c85d

profiles/utilities.py
```diff
@@ -15,3 +17,10 @@ def generate_random_string(length: int, chars: str = string.ascii_letters + stri
 
 def remove_none_values(items: dict) -> dict:
     return dict(filter(lambda item: item[1] is not None, items.items()))
+
+
+def string_to_name(name: str) -> Name:
```
giorgiosironi: I understand this is temporary

profiles/models.py
```diff
@@ -21,12 +23,40 @@ def __repr__(self) -> str:
...
+    def __ne__(self, other: Any) -> bool:
+        return not self.__eq__(other)
+
+
 class Profile(db.Model):
     id = db.Column(db.String(ID_LENGTH), primary_key=True)
-    name = db.Column(db.Text(), nullable=False)
+    name = composite(Name, '_given_names', '_family_name')
+    _given_names = db.Column(db.Text(), name='given_names', nullable=False)
+    _family_name = db.Column(db.Text(), name='family_name')
```
giorgiosironi: do we need to store parts of the name separately? They may not be available when using non-ORCID data sources

giorgiosironi: also, we may need to store `index` values like https://github.com/elifesciences/api-raml/pull/192/files#diff-6cbf65dc93ac1e6649327ba4a25bc550R15 directly as a computed field to allow for (fast) ordering (or what that field is used for?)

giorgiosironi: It is true we lose information, but it can be retrieved from the API and this is our data model's, not ORCID's

thewilkybarkid: Hmm, had thought that sorting by family then given names would work, but that would actually produce different results (when there is no family name).

In the future I envisaged a requirement for an 'address' name (ie Hello Chris in an email), but we can always cross that bridge when it comes.

profiles/utilities.py
thewilkybarkid: Only if we want to consider successful auth then immediate API failure as a Profiles auth failure. (Otherwise we only have access to a single-string name.)

PR #30, opened 2017-09-25
ELPP-2984 Add API

thewilkybarkid:


profiles/api/api.py
```diff
@@ -0,0 +1,54 @@
...
+
+        if page < 1:
+            raise BadRequest('Page less than 1')
+        elif page != request.args.get('page'):
+            raise BadRequest('Invalid page')
+
+        if per_page < 1 or per_page > 100:
+            raise BadRequest('Per page out of range')
+        elif page != request.args.get('per-page'):
+            raise BadRequest('Invalid per page')
```
thewilkybarkid: This all feels a bit verbose.

profiles/serializer/normalizer.py
```diff
@@ -0,0 +1,38 @@
+from collections import OrderedDict
+from functools import singledispatch
+from typing import Any
+
+from profiles.models import Profile
+
+
+@singledispatch
+def normalize(value: Any) -> str:
```
thewilkybarkid: Not sure of a good Python pattern for normalising/denormalising JSON to/from an object.

profiles/api/api.py
giorgiosironi: There is probably some validation library for Flask out there, but is it worth it for a single endpoint with parameters?

profiles/api/api.py
```diff
@@ -0,0 +1,54 @@
+import json
```
giorgiosironi: generic name for a `api` module, could have been `profiles.api.profiles` (which is even more confusing)

profiles/api/api.py
```diff
@@ -0,0 +1,54 @@
...
+        return response
+
+    @blueprint.route('/profiles/<id>')
+    def _get(id: str) -> Response:
+        try:
+            profile = profiles.get(id)
+        except ProfileNotFound as exception:
+            raise NotFound(str(exception)) from exception
+
+        response = make_response(json.dumps(profile, default=normalize))
```
giorgiosironi: a little worried that this will print `{"field": "<SomeClass>0x1234"}` sooner or later, but there are tests for that

profiles/repositories.py
```diff
@@ -92,5 +96,8 @@ def next_id(self) -> str:
 
         return profile_id
 
+    def list(self, limit: int = None, offset: int = 0) -> List[Profile]:
+        return self.db.session.query(Profile).limit(limit).offset(offset).all()
```
giorgiosironi: is there a default ordering? making it explicit here?

profiles/serializer/normalizer.py
giorgiosironi: I'm in the dark too

profiles/serializer/normalizer.py
```diff
@@ -0,0 +1,38 @@
...
+    return data
+
+
+@normalize_snippet.register(Profile)
+def normalize_profile_snippet(profile: Profile) -> dict:
+    data = OrderedDict([
+        ('id', profile.id),
+        ('name', OrderedDict([
+            ('preferred', profile.name),
+            ('index', profile.name),
```
giorgiosironi: will probably be updated later with another field?

test/test_api.py
```diff
@@ -0,0 +1,132 @@
+import json
+
+from flask.testing import FlaskClient
+
+from profiles.models import Profile, db
+
+
+def test_empty_list_of_profile(test_client: FlaskClient) -> None:
```
giorgiosironi: profile`s`?

profiles/api/api.py
thewilkybarkid: Could go the whole hog and call it `profiles.api.profiles_api` in case it implements a second endpoint (unlikely?).

Though maybe `profiles.api` isn't the best name.

profiles/repositories.py
thewilkybarkid: Depends on #29.

profiles/serializer/normalizer.py
thewilkybarkid: Yep, after #29.

test/test_api.py
thewilkybarkid: 👓

profiles/repositories.py
```diff
@@ -92,5 +96,15 @@ def next_id(self) -> str:
 
         return profile_id
 
+    def list(self, limit: int = None, offset: int = 0, desc: bool = True) -> List[Profile]:
+        query = self.db.session.query(Profile)
+
+        if desc:
+            query = query.order_by(Profile._index_name.desc())  # pylint: disable=protected-access
```
thewilkybarkid: There has to be a better way than this?

giorgiosironi: doesn't look like an SQLAlchemy problem as much as accessing a non-public class field. This could be exposed as a static method from the `Profile` class so that only a few selected orderings are available?
```
@staticmethod
def asc():
    return _index_name.asc()

query.order_by(Profile.asc()) # yeah
query.order_by(Profile._other_field.asc()) # you can't
```

thewilkybarkid: I thought `query = query.order_by(Profile.name.index.desc())` would work, but get `Neither 'property' object nor 'Comparator' object associated with Profile.name has an attribute 'index'`.

For minute we only order by index name, so that might be ok.

giorgiosironi:
Depends on https://github.com/elifesciences/builder-private/pull/57 but then we could merge to try it.

profiles/repositories.py
seanwiseman: @thewilkybarkid could you look over the changes I have committed? as you are the author of the PR I can't select you as a reviewer.

profiles/models.py
```diff
@@ -101,6 +111,18 @@ def __init__(self, affiliation_id: str, address: Address, organisation: str, sta
...
+        return [self.department, self.organisation]
+
+    def is_current(self) -> bool:
+        time_now = pendulum.timezone('utc').convert(datetime.now())
+
+        if self.starts <= time_now:
+            if not self.ends or self.ends > time_now:
+                return True
+
+        return False
```
thewilkybarkid: Could be:

```python
starts = pendulum.instance(self.starts)
ends = pendulum.instance(self.ends) if self.ends else None

return starts.is_past() and (not ends or ends.is_future())
```

profiles/models.py
```diff
@@ -60,7 +60,17 @@ def __init__(self, country: Country, city: str, region: str = None) -> None:
...
+        return self.country, self.city, self.region
+
+    def get_formatted(self) -> List[str]:
+        return [self.city, self.region, self.country.alpha2]
+
+    def get_components(self) -> Dict:
+        return {
+            "locality": [self.city],
+            "area": [self.region],
+            "country": self.country.name
```
thewilkybarkid: `get_formatted()` and `get_components()` are leaking JSON representation details into the model, we should keep them in the normalizer.

thewilkybarkid: Same with `Affiliation. get_name_list()`.

profiles/serializer/normalizer.py
```diff
@@ -0,0 +1,60 @@
...
+def normalize_profile(profile: Profile) -> dict:
+    data = normalize_profile_snippet(profile)
+    data['emailAddresses'] = [normalize_snippet(email) for email in profile.email_addresses]
+    data['affiliations'] = [normalize_snippet(aff) for aff in profile.get_affiliations()]
+
+    return data
+
+
+@normalize_snippet.register(Affiliation)
+def normalize_affiliation_snippet(affiliation: Affiliation) -> dict:
```
thewilkybarkid: Snippets only exist in the top level (full Profile vs Profile snippet), so I'd use `normalize()`.

thewilkybarkid: Same for email addresses.

profiles/serializer/normalizer.py
```diff
@@ -0,0 +1,60 @@
...
+@singledispatch
+def normalize_snippet(value: Any) -> str:
+    return str(value)
+
+
+@normalize.register(Profile)
+def normalize_profile(profile: Profile) -> dict:
+    data = normalize_profile_snippet(profile)
+    data['emailAddresses'] = [normalize_snippet(email) for email in profile.email_addresses]
+    data['affiliations'] = [normalize_snippet(aff) for aff in profile.get_affiliations()]
```
thewilkybarkid: We'll need to ignore email addresses and affiliations that are `restricted`. (We might want them in the future, but only for API users with permission.)

test/test_api.py
```diff
@@ -0,0 +1,235 @@
...
+
+    db.session.commit()
+
+    response = test_client.get('/profiles/a1b2c3d4')
+    response_data = json.loads(response.data.decode('UTF-8'))
+
+    assert validate_json(response_data, schema_name='profile.v1') is True
+
+
+def test_get_list_of_profiles_response_against_json_schema(test_client: FlaskClient) -> None:
```
thewilkybarkid: Ideally all the above tests would validate instead of having separate tests.

test/serializer/test_normalize.py
```diff
@@ -0,0 +1,162 @@
...
+        'emailAddresses': [],
+        'affiliations': [
+            {
+                "name": ["Dep", "Org2"],
+                "address": {
+                    "formatted": ["City2", "Region2", "GB"],
+                    "components": {
+                        "locality": ["City2"],
+                        "area": ["Region2"],
+                        "country": "United Kingdom of Great Britain and Northern Ireland"
```
thewilkybarkid: Wonder if we can get a short name out of the library (ie `United Kingdom`)...

profiles/models.py
seanwiseman: `is_past` and `is_future` are handy, less manual work. Changed.

test/models/test_affiliation.py
```diff
@@ -87,3 +87,49 @@ def test_it_may_be_restricted():
...
+    address = Address(countries.get('gb'), 'City')
+    affiliation = Affiliation('1', address=address, organisation='Org',
+                              department='Dep', starts=start_date)
+
+    assert affiliation.is_current() is True
+
+
+def test_it_can_detect_if_current_with_ends_date():
+    start_date = datetime(2017, 1, 1, 1, 0, 0, tzinfo=timezone(timedelta(hours=1)))
+    ends_date = datetime(2100, 1, 1, 1, 0, 0, tzinfo=timezone(timedelta(hours=1)))
```
thewilkybarkid: I've been using `@freeze_time()` where-ever tests have meaningful dates. Means you don't have to set such things.

profiles/serializer/normalizer.py
seanwiseman: Ok these have been moved to normalise now.

profiles/models.py
seanwiseman: OK agreed. Now all representation is in the normalize body:
```
@normalize.register(Affiliation)
def normalize_affiliation(affiliation: Affiliation) -> dict:
    return {
        "name": [affiliation.department, affiliation.organisation],
        "address": {
            "formatted": [
                affiliation.address.city,
                affiliation.address.region,
                affiliation.address.country.alpha2,
            ],
            "components": {
                "locality": [affiliation.address.city],
                "area": [affiliation.address.region],
                "country": affiliation.address.country.name
            }
        }
    }
```

test/test_api.py
seanwiseman: Ok done.

test/models/test_affiliation.py
seanwiseman: I can't use `@freeze_time()` when also using `Affiliation.is_current()` as the `pendulum.instance.is_past()` uses `< datetime.datetime.now()` under the hood. Meaning it never thinks something is in the past.

If there are any other places I can use it I will add it.

test/serializer/test_normalize.py
seanwiseman: The only non abbreviated or numeric fields are `name` (currently being used) or `apolitical_name` which seems to be Identical in the library:
```
Country(u"United Kingdom of Great Britain and Northern Ireland",
            "GB", "GBR", "826",
            u"United Kingdom of Great Britain and Northern Ireland"),
```
Just seems that the entry for the UK is rather wordy.

thewilkybarkid: Hmm, we might need to have our own set of names.

profiles/serializer/normalizer.py
seanwiseman: I have now added default functionality to filter out restricted email addresses and affiliations unless you flag to included them, which could be tied into another endpoint for users with said permission in the future.

profiles/serializer/normalizer.py
```diff
@@ -17,25 +17,33 @@ def normalize_snippet(value: Any) -> str:
...
     return {
-        "name": affiliation.get_name_list(),
+        "name": [affiliation.department, affiliation.organisation],
         "address": {
-            "formatted": affiliation.address.get_formatted(),
-            "components": affiliation.address.get_components()
+            "formatted": [
+                affiliation.address.city,
+                affiliation.address.region,
+                affiliation.address.country.alpha2,
```
thewilkybarkid: `name` rather than `alpha2`.

profiles/models.py
```diff
@@ -187,6 +193,12 @@ def add_email_address(self, email: str, primary: bool = False,
 
         self.email_addresses.reorder()
 
+    def get_email_addresses(self, include_restricted: bool = False) -> List[str]:
+        if include_restricted:
+            return [email for email in self.email_addresses]
```
thewilkybarkid: `return self.email_addresses`?

profiles/utilities.py
```diff
@@ -1,8 +1,12 @@
+import json
+import os
 import random
 import string
 from datetime import datetime
 
+from jsonschema import SchemaError, ValidationError, validate
 import pendulum
+from profiles.exceptions import SchemaNotFound
```
thewilkybarkid: Needs a space above

profiles/utilities.py
```diff
@@ -22,3 +26,21 @@ def guess_index_name(name: str) -> str:
 
 def remove_none_values(items: dict) -> dict:
     return dict(filter(lambda item: item[1] is not None, items.items()))
+
+
+def validate_json(data: dict, schema_name: str):
+    schema_dir = 'test/schema'  # Will be replaced with elife_api.schema_directory
```
thewilkybarkid: Maybe make this an explicit TODO?

thewilkybarkid: In fact, we can load them now can't we?

requirements.txt
```diff
@@ -44,3 +45,5 @@ urllib3==1.22
 uWSGI==2.0.15
 Werkzeug==0.12.2
 wrapt==1.10.11
+
+# git+https://github.com/elifesciences/api-raml.git#egg=elife_api
```
thewilkybarkid: Needs removing.

test/models/test_affiliation.py
thewilkybarkid: Ah yeah, forget how modules work! Could mock it to return the same value?

profiles/utilities.py
seanwiseman: Yes was going to wait until the `api-validator` was complete but no real reason not to use it for the schemas now. I have included it as a a dependancy now and amended the temporary validation function for the time being.

requirements.txt
seanwiseman: Have now replaced with `api-validator-python` url.

`-e git+https://github.com/elifesciences/api-validator-python.git#egg=api-validator-python`

Note the use of the `-e` flag for editable, without this it does not unpack the non Python files into the site-packages directory as by default libraries are set to non editable.

test/models/test_affiliation.py
seanwiseman: I have used `@freeze_time()` where I am only looking for future dates (no past dates to check, so no problem) but to get it working in the use case mentioned I would need to to `@freeze_time()` then mock out the pendulum return value to be less than the `@freeze_time()` value as `is_past()` only accepts `< datetime.now()` as `True`. You trade off an extra variable being declared in the function with an extra decorator and a function arg for the mock. I think anywhere else is fine, but if we are actually testing `is_current()` its easier to just declare the variable.

thewilkybarkid:
👍

PR #31, opened 2017-09-25
Handle private names

thewilkybarkid:
I assumed that no `name` would be provided if it were private (and so would fail), turns out that it's an empty string, which we've now persisted.

~Instead of hard-failing, this persists a clear fallback value.~

### DIFF
--- a/profiles/api/oauth2.py
+++ b/profiles/api/oauth2.py
@@ -127,10 +127,10 @@ def create_blueprint(orcid: Dict[str, str], clients: Clients, profiles: Profiles
 
         try:
             profile = profiles.get_by_orcid(json_data['orcid'])
-            if json_data.get('name', ''):
+            if json_data['name']:
                 profile.name = Name(json_data['name'])
         except ProfileNotFound:
-            if not json_data.get('name', ''):
+            if not json_data['name']:
                 raise InvalidRequest('No name visible')
             profile = Profile(profiles.next_id(), Name(json_data['name']), json_data['orcid'])
             profiles.add(profile)
--- a/profiles/api/oauth2.py
+++ b/profiles/api/oauth2.py
@@ -118,21 +118,20 @@ def create_blueprint(orcid: Dict[str, str], clients: Clients, profiles: Profiles
 
         json_data = json.loads(response.text)
 
-        if 'access_token' not in json_data:
-            raise ValueError('No access_token')
-        elif 'expires_in' not in json_data:
-            raise ValueError('No expires_in')
-        elif json_data.get('token_type').lower() != 'bearer':
-            raise ValueError('Got token_type {}, expected Bearer'.format(
-                json_data.get('token_type')))
-
         json_data = {key: json_data[key] for key in
                      ['access_token', 'expires_in', 'name', 'orcid', 'token_type']}
 
+        if json_data.get('token_type').lower() != 'bearer':
+            raise ValueError('Got token_type {}, expected Bearer'.format(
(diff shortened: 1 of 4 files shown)


thewilkybarkid:
In fact, I remember the spec being to reject profiles without a visible name.

profiles/api/oauth2.py
```diff
@@ -118,21 +118,20 @@ def _token() -> Response:
...
+                json_data.get('token_type')))
+
         try:
             profile = profiles.get_by_orcid(json_data['orcid'])
-            profile.name = Name(json_data['name'])
+            if json_data.get('name', ''):
+                profile.name = Name(json_data['name'])
         except ProfileNotFound:
+            if not json_data.get('name', ''):
+                raise InvalidRequest('No name visible')
```
thewilkybarkid: I'm guessing this is the right type of error. Journal will need to be able to display a friendly error message.

profiles/api/oauth2.py
```diff
@@ -118,21 +118,20 @@ def _token() -> Response:
...
+                json_data.get('token_type')))
+
         try:
             profile = profiles.get_by_orcid(json_data['orcid'])
-            profile.name = Name(json_data['name'])
+            if json_data['name']:
+                profile.name = Name(json_data['name'])
         except ProfileNotFound:
+            if not json_data['name']:
+                raise InvalidRequest('No name visible')
```
thewilkybarkid: I'm guessing this is the right type of error.

Journal will need to be able to display a friendly error message from this.

profiles/api/oauth2.py
```diff
@@ -118,21 +118,20 @@ def _token() -> Response:
...
 
-        if 'access_token' not in json_data:
-            raise ValueError('No access_token')
-        elif 'expires_in' not in json_data:
-            raise ValueError('No expires_in')
-        elif json_data.get('token_type').lower() != 'bearer':
-            raise ValueError('Got token_type {}, expected Bearer'.format(
-                json_data.get('token_type')))
-
         json_data = {key: json_data[key] for key in
```
giorgiosironi: doesn't this raise `KeyError` when the field is not there?

thewilkybarkid: Yep, which is fine (ORCID have given us a broken response). The `ValueError`s above were redundant.

test/test_oauth2.py
```diff
@@ -280,6 +280,44 @@ def test_it_updates_a_profile_when_exchanging(test_client: FlaskClient) -> None:
...
+                                          'client_secret': 'client_secret',
+                                          'redirect_uri': 'http://www.example.com/client/redirect',
+                                          'grant_type': 'authorization_code', 'code': '1234'})
+
+    assert response.status_code == 400
+    assert response.headers.get('Content-Type') == 'application/json'
+
+
+def test_it_ignores_now_private_names_when_exchanging(test_client: FlaskClient) -> None:
+    original_profile = Profile('a1b2c3d4', Name('Foo Bar'), '0000-0002-1825-0097')
```
giorgiosironi: good setup that doesn't go through the original exchange but creates a local `Profile` only

profiles/api/oauth2.py
giorgiosironi: Is this reported in the body? If so it can be checked in the tests?

thewilkybarkid: It is. Never a fan of parsing error messages, but we don't have access to a code (unless we make one up).

PR #32, opened 2017-09-27
Fix test finding

thewilkybarkid:
#27 added tests in subdirectories, but they aren't being found in CI and subsequent changes have broken some.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,4 +1,4 @@
 [pytest]
 addopts = --disable-socket
 cache_dir = build/pytest
-python_files = test/test_*.py
+python_files = test/**/test_*.py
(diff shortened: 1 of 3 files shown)


pytest.ini
```diff
@@ -1,4 +1,4 @@
 [pytest]
 addopts = --disable-socket
 cache_dir = build/pytest
-python_files = test/test_*.py
+python_files = test_*.py
```
giorgiosironi: since this is Python 3.5, this could probably be `test/**/test_*.py`

giorgiosironi: https://stackoverflow.com/questions/2186525/use-a-glob-to-find-files-recursively-in-python

thewilkybarkid: I tried that first, but Pytest uses `path.fnmatch` rather than `glob.glob`, so it only results in the contents of 1 directory deep being found. Planning to open a PR.

PR #33, opened 2017-09-29
Introducing clear command

giorgiosironi:
Useful for cleaning a test environment from data. And in production, if you also like to live dangerously

Not sure about the name of the command.


profiles/commands.py
```diff
@@ -0,0 +1,16 @@
+from flask_script import Command
+
+
+class ClearCommand(Command):
+    def __init__(self, *repositories):
```
thewilkybarkid: Can we add a `CanBeCleared` type? We might not just want to clear repositories (thinking caches).

manage.py
```diff
@@ -33,5 +34,7 @@ def make_shell_context() -> dict:
 manager.add_command('runserver', Server())
 manager.add_command('shell', Shell(make_context=make_shell_context))
 
+[manager.add_command('delete', c) for c in app.commands]
```
thewilkybarkid: Doesn't the command need to define what its name is? (Aren't all commands going to be set as `delete`?)

thewilkybarkid:
Probably ok to not react to profiles being deleted (thinking emitting events, unregistering webhooks).

profiles/commands.py
```diff
@@ -1,6 +1,24 @@
+from flask_script import Command
 from profiles.models import Name, Profile
 
 
+class ClearCommand(Command):
```
thewilkybarkid: This is mixing CLI commands with domain commands.

profiles/commands.py
```diff
@@ -1,6 +1,24 @@
...
+    NAME = 'clear'
+
+    def __init__(self, *repositories):
+        super(ClearCommand, self).__init__()
+        self.repositories = repositories
+
+    # weird decoration made by the superclass that wraps this method
+    # not going to lose sleep over this for now,
+    # let's see whether we get more commands
+    # and we keep flask_script in the long term
```
thewilkybarkid: Reading https://flask-script.readthedocs.io/en/latest/ we should probably not.

giorgiosironi:
Updated (also with the `develop` branch)

test/repositories/sql_alchemy/test_profiles.py
```diff
@@ -71,3 +71,15 @@ def id_generator():
 
     with pytest.raises(RuntimeError):
         assert profiles.next_id()
+
+
+def test_it_clears_profiles():
+    def id_generator():
+        return '11111111'
```
thewilkybarkid: Not needed (`profiles.next_id()` isn't called).

giorgiosironi:
Tried to put `manage.py` under pylint but would require a uwsgi configuration change due to variable names, not doing it now.

profiles/cli.py
```diff
@@ -1,10 +1,12 @@
 from flask_script import Command
 
+from profiles.types import CanBeCleared
+
 
 class ClearCommand(Command):
     NAME = 'clear'
 
-    def __init__(self, *repositories):
+    def __init__(self, *repositories: CanBeCleared) -> None:
```
giorgiosironi: this works for argument "lists" too, nice

PR #37, opened 2017-10-05
Updated orcid-dummy to: null

elife-alfred-user:
I have run http://ci--alfred.elifesciences.org/job/dependencies-profiles-update-orcid-dummy/6/ which resulted in this pull request.


giorgiosironi:
The title could use some work.

PR #42, opened 2017-10-09
ELPP-2775 Fetch affiliations

thewilkybarkid:


profiles/models.py
```diff
@@ -51,12 +52,53 @@ def __ne__(self, other: Any) -> bool:
...
+    region = db.Column(db.Text())
+    _country = db.Column(db.String(2), name='country', nullable=False)
+    restricted = db.Column(db.Boolean(), nullable=False)
+    profile_id = db.Column(db.String(ID_LENGTH), db.ForeignKey('profile.id'))
+    profile = db.relationship('Profile', back_populates='affiliations')
+    position = db.Column(db.Integer())
+
+    # pylint: disable=too-many-arguments
+    def __init__(self, country: Country, organisation: str, department: str = None,
+                 city: str = None, region: str = None, restricted: bool = False) -> None:
```
thewilkybarkid: More matches ORCID's structure than our API's. `country`, `region` and `city` could form an `Address`.

profiles/models.py
```diff
@@ -51,12 +52,53 @@ def __ne__(self, other: Any) -> bool:
...
+        self.department = department
+        self.organisation = organisation
+        self.city = city
+        self.region = region
+        self.country = country
+        self._country = country.alpha2
```
thewilkybarkid: Could make a `Country` SQLAlchemy type?

profiles/models.py
```diff
@@ -51,12 +52,53 @@ def __ne__(self, other: Any) -> bool:
         return not self.__eq__(other)
 
 
+class Affiliation(db.Model):
+    _id = db.Column(db.Integer, name='id', primary_key=True)
```
thewilkybarkid: Nearest thing to an ID that the ORCID API appears to return is a `put-code`, so comparing values instead. Need to look more into it.

thewilkybarkid:
Need to remove past employments.

thewilkybarkid:
Need to find out what happens with start/end date in the future. We (probably) want affiliations to appear/disappear at the right time. Is it possible? (I think so.) Do we get a web hook notification at that point. (I'm guessing not.)

If so, we'll have to store the start/end date and filter them as appropriate. (Maybe just store all affiliations rather than current ones?)

profiles/models.py
giorgiosironi: if this sort of storage/structural detail (`alpha2`) is being spread over domain classes, yes

profiles/models.py
```diff
@@ -66,6 +108,23 @@ def __init__(self, profile_id: str, name: Name, orcid: str = None) -> None:
...
+            if existing_affiliation == affiliation:
+                existing_affiliation.restricted = affiliation.restricted
+                if position != existing_affiliation.position:
+                    self.affiliations.remove(existing_affiliation)
+                    self.affiliations.insert(position, existing_affiliation)
+                    self.affiliations.reorder()
+                return
+
+        self.affiliations.insert(position, affiliation)
+        self.email_addresses.reorder()
```
giorgiosironi: need to reorder the `email_addresses`? :thinking:

profiles/commands.py
```diff
@@ -1,8 +1,12 @@
-from profiles.models import Name, Profile
+from iso3166 import countries
+from pendulum import create as date
+
+from profiles.models import Affiliation, Name, Profile
 
 
 def update_profile_from_orcid_record(profile: Profile, orcid_record: dict) -> None:
     _update_name_from_orcid_record(profile, orcid_record)
+    _update_affiliations_from_orcid_record(profile, orcid_record)
```
giorgiosironi: these are starting to become potentially independent updates e.g. if it gets too slow we could do some of them in background

profiles/commands.py
```diff
@@ -15,6 +19,34 @@ def _update_name_from_orcid_record(profile: Profile, orcid_record: dict) -> None
...
 
 
+def _update_affiliations_from_orcid_record(profile: Profile, orcid_record: dict) -> None:
+    def create_affiliation(properties: dict) -> Affiliation:
+        organization = properties['organization']
+        address = organization['address']
+
+        return Affiliation(countries.get(address['country']), organization['name'],
+                           properties.get('department-name'), address.get('city'),
+                           address.get('region'), properties['visibility'] != 'PUBLIC')
```
giorgiosironi: `PUBLIC` is calling for a constant

profiles/commands.py
thewilkybarkid: They mostly will be, it's just the first authentication that requires it (we won't need to do it subsequent times as we'll be using the webhook).

thewilkybarkid:
Confirmed, you can have future dated employments appear (both start and end date). So I think we'll need to store the whole history. I think we'll need it in the future anyhow.

thewilkybarkid:
The `put-code` does look to be a unique ID (looks like an auto-increment), might be safe to use (and would simplify updates).

profiles/commands.py
```diff
@@ -15,6 +20,34 @@ def _update_name_from_orcid_record(profile: Profile, orcid_record: dict) -> None
                             '{}, {}'.format(family_name, given_name))
 
 
+def _update_affiliations_from_orcid_record(profile: Profile, orcid_record: dict) -> None:
```
giorgiosironi: https://github.com/elifesciences/orcid-dummy/pull/4 provides data for this in end2end

thewilkybarkid:
This should now fetch and synchronise the full history.

profiles/commands.py
```diff
@@ -15,6 +20,38 @@ def _update_name_from_orcid_record(profile: Profile, orcid_record: dict) -> None
...
+    orcid_affiliations = orcid_record.get('activities-summary', {}).get('employments', {}).get(
+        'employment-summary', {})
+
+    for index, orcid_affiliation in enumerate(orcid_affiliations):
+        organization = orcid_affiliation['organization']
+        address = organization['address']
+
+        affiliation = Affiliation(str(orcid_affiliation['put-code']),
```
giorgiosironi: named parameters could be useful in showing which field each value corresponds to

profiles/commands.py
```diff
@@ -15,6 +20,38 @@ def _update_name_from_orcid_record(profile: Profile, orcid_record: dict) -> None
...
+                                  countries.get(address['country']), organization['name'],
+                                  date(**{k: int(v) for k, v in
```
giorgiosironi: what a coincidence that `pendulum.create` takes exactly the same arguments as the fields contained in `start-date` or `end-date`? Seems too magic?

thewilkybarkid: Yeah, had wondered that in 65e473cfea5708c6de6452bfbf28366360188301...

profiles/commands.py
```diff
@@ -15,6 +20,38 @@ def _update_name_from_orcid_record(profile: Profile, orcid_record: dict) -> None
...
+                                          orcid_affiliation['start-date'].items()}),
+                                  orcid_affiliation.get('department-name'), address.get('city'),
+                                  address.get('region'),
+                                  date(**{k: int(v) for k, v in
```
giorgiosironi: conceptually if this is a date, it has no hour/minutes/seconds. I guess there is no correct data type to store `2017-01-01`?

profiles/commands.py
```diff
@@ -15,6 +20,38 @@ def _update_name_from_orcid_record(profile: Profile, orcid_record: dict) -> None
...
+                                          orcid_affiliation['end-date'].items()}, hour=23,
+                                       minute=59, second=59) if orcid_affiliation.get('end-date')
+                                  else None,
+                                  orcid_affiliation['visibility'] != VISIBILITY_PUBLIC)
+        profile.add_affiliation(affiliation, index)
+
+    for affiliation in profile.affiliations:
+        found = False
+        for orcid_affiliation in orcid_affiliations:
+            if str(orcid_affiliation['put-code']) == affiliation.id:
```
giorgiosironi: I imagine `put-code`s are unique per-ORCID platform? All records shared a common space?

thewilkybarkid: I _think_ so; alternatively we could continue to generate a local ID and put a unique constraint on their ID + profile ID.

giorgiosironi: https://members.orcid.org/api/tutorial/reading-xml says
```
Put codes are unique within the type of item they refer to. Using this example, only one work in the ORCID Registry will have the put code 733535, though 733535 may also used as a put code for a funding item or a keyword.
```
which is good enough for this use case

profiles/commands.py
```diff
@@ -15,6 +20,38 @@ def _update_name_from_orcid_record(profile: Profile, orcid_record: dict) -> None
...
+                                  orcid_affiliation.get('department-name'), address.get('city'),
+                                  address.get('region'),
+                                  date(**{k: int(v['value']) for k, v in
+                                          orcid_affiliation['end-date'].items()}, hour=23,
+                                       minute=59, second=59) if orcid_affiliation.get('end-date')
+                                  else None,
+                                  orcid_affiliation['visibility'] != VISIBILITY_PUBLIC)
+        profile.add_affiliation(affiliation, index)
+
+    for affiliation in profile.affiliations:
```
giorgiosironi: it looks like this iterates on `profile.affiliations` to diff it from `orcid_affiliations`, but two lines ago everything from `orcid_affiliatons` was added to `profile`?

thewilkybarkid: To find any entries that have been removed.

profiles/models.py
```diff
@@ -51,12 +53,48 @@ def __ne__(self, other: Any) -> bool:
...
+    country = db.Column(ISO3166Country, nullable=False)
+    starts = db.Column(UTCDateTime, nullable=False)
+    ends = db.Column(UTCDateTime)
+    restricted = db.Column(db.Boolean(), nullable=False)
+    profile_id = db.Column(db.String(ID_LENGTH), db.ForeignKey('profile.id'))
+    profile = db.relationship('Profile', back_populates='affiliations')
+    position = db.Column(db.Integer())
+
+    # pylint: disable=too-many-arguments
+    def __init__(self, affiliation_id: str, country: Country, organisation: str, starts: datetime,
```
giorgiosironi: perhaps an Address object, as the API suggests, could be useful; not sure if it's a nightmare to map onto the database

giorgiosironi: (as a Value Object decomposed into multiple fields on the owning entity, only advanced ORMs support it http://docs.oracle.com/javaee/6/api/javax/persistence/Embeddable.html)

thewilkybarkid: Done already for the name: https://github.com/elifesciences/profiles/blob/develop/profiles/models.py#L56-L58

(Not as smooth as it could be.)

giorgiosironi: Not bad, boilerplate is limited to one line per decomposed hidden field (which is needed to decide on the column type anyway) plus the decomposition method

profiles/commands.py
giorgiosironi: I see how that works now, but if we build a `added_ids` `Set` the complexity of this can be reduced to `if affiliation.id in added_ids` (both in terms of computation and readability)

thewilkybarkid: 👍

profiles/models.py
thewilkybarkid: Our API does also have an 'address' concept.

PR #45, opened 2017-10-12
Add property based tests to the test suite

seanwiseman:
I started by applying these to the tests in the client suite as they were mainly just dependant on string values and were fairly concise. If it is agreed that this testing approach is advantageous for Profiles I can continue and work through the remaining suite adding them where suitable.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+/.hypothesis
 /app.cfg
 /build/*
 /clients.yaml
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-/.hypothesis
 /app.cfg
 /build/*
 /clients.yaml
--- a/test/clients/test_clients.py
+++ b/test/clients/test_clients.py
@@ -1,15 +1,12 @@
 from hypothesis import given
-from hypothesis.searchstrategy.lazy import LazyStrategy
 from hypothesis.strategies import lists, text
+
 import pytest
+
 from profiles.clients import Client, Clients
 
 
-def list_of_str(length: int) -> LazyStrategy:
-    return lists(text(), max_size=length, min_size=length)
-
-
-@given(list_of_str(4), list_of_str(4))
+@given(lists(text(), max_size=4, min_size=4), lists(text(), max_size=4, min_size=4))
 def test_it_contains_clients(client1_args, client2_args):
     clients = Clients(Client(*client1_args), Client(*client2_args))
 
@@ -20,7 +17,7 @@ def test_it_contains_clients(client1_args, client2_args):
 def test_it_finds_clients(id_base):
     client_id1 = id_base + '1'
     client_id2 = id_base + '2'
-    unknown_client_id = id_base + 'unkwown'
+    unknown_client_id = id_base + 'unknown'
 
     client1 = Client('name1', client_id1, 'client_secret1', 'redirect_uri1')
     client2 = Client('name2', client_id2, 'client_secret2', 'redirect_uri2')
--- a/test/clients/test_clients.py
+++ b/test/clients/test_clients.py
@@ -1,25 +1,35 @@
+from hypothesis import given
+from hypothesis.searchstrategy.lazy import LazyStrategy
+from hypothesis.strategies import lists, text
 import pytest
-
 from profiles.clients import Client, Clients
 
 
-def test_it_contains_clients():
-    client1 = Client('name1', 'client_id1', 'client_secret1', 'redirect_uri1')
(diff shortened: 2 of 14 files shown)


.gitignore
```diff
@@ -1,3 +1,4 @@
+/.hypothesis
```
thewilkybarkid: Can we configure this to go under `build/`?

test/clients/test_clients.py
```diff
@@ -1,25 +1,35 @@
+from hypothesis import given
+from hypothesis.searchstrategy.lazy import LazyStrategy
+from hypothesis.strategies import lists, text
 import pytest
-
```
thewilkybarkid: Been following having imports as separate groups (though not found a linter than can enforce it). (`test_client.py` should be separated too.)

test/clients/test_clients.py
```diff
@@ -1,25 +1,35 @@
...
 
 
-def test_it_finds_clients():
-    client1 = Client('name1', 'client_id1', 'client_secret1', 'redirect_uri1')
-    client2 = Client('name2', 'client_id2', 'client_secret2', 'redirect_uri2')
+@given(text(min_size=1))
+def test_it_finds_clients(id_base):
+    client_id1 = id_base + '1'
+    client_id2 = id_base + '2'
+    unknown_client_id = id_base + 'unkwown'
```
thewilkybarkid: `unkwown` 🙃

test/clients/test_clients.py
```diff
@@ -1,25 +1,35 @@
...
 import pytest
-
 from profiles.clients import Client, Clients
 
 
-def test_it_contains_clients():
-    client1 = Client('name1', 'client_id1', 'client_secret1', 'redirect_uri1')
-    client2 = Client('name2', 'client_id2', 'client_secret2', 'redirect_uri2')
-    clients = Clients(client1, client2)
+def list_of_str(length: int) -> LazyStrategy:
```
thewilkybarkid: Nice and simple to do.

Might be worth moving this elsewhere (`conftest.py`?). Maybe rename to `list_of_text` (as it's wrapping `text()`), or maybe even make it generic (`list_of(text(), 4)`) if that's possible?

.gitignore
seanwiseman: Sure. Two options for this, you can either set this in a .settings object along with any other required values, or in an HYPOTHESIS_DATABASE_FILE env variable. Whats your view?

thewilkybarkid: The former is preferable, we do the same for pytest (https://github.com/elifesciences/profiles/blob/develop/pytest.ini#L3).

test/clients/test_clients.py
seanwiseman: OK, changed.

test/clients/test_clients.py
seanwiseman: That is the next level after 'unknown'... now changed.

test/clients/test_clients.py
seanwiseman: Upon further investigation, the preferred pattern seems to be just declaring the '@given' arguments explicitly rather than through helper functions. The '@fixture's play nice alongside '@given' but not so much with, which is fine as long as you are being explicit.

```
@given(lists(text(), max_size=4, min_size=4), lists(text(), max_size=4, min_size=4))
def test_it_contains_clients(client1_args, client2_args):
```

thewilkybarkid: Fair enough. :+1:

test/conftest.py
```diff
@@ -16,16 +17,19 @@
...
 TEST_DATABASE_NAME = 'test.db'
-TEST_DATABASE_PATH = os.path.dirname(os.path.realpath(__file__)) + '/../build/' + TEST_DATABASE_NAME
+TEST_DATABASE_PATH = BUILD_PATH + TEST_DATABASE_NAME
 TEST_DATABASE_URI = 'sqlite:///' + TEST_DATABASE_PATH
 
 logging.disable(logging.CRITICAL)
 
-hyp_settings.register_profile("default", hyp_settings(database_file='build/hypothesis'))
+set_hypothesis_home_dir(BUILD_PATH + 'hypothesis/home')
+hyp_settings.register_profile('default', hyp_settings(database_file=BUILD_PATH + 'hypothesis/db'))
```
thewilkybarkid: Has to be absolute otherwise is relative to the test file being executed.

seanwiseman: 👍🏻

seanwiseman:
I am currently going through and adding more property based tests where appropriate.

seanwiseman:
I have added the remaining tests. I have tried not to touch tests that perform similar tasks or setup to ones already covered.

PR #46, opened 2017-10-26
Gracefully fail on problems reading ORCID data

thewilkybarkid:
For bugs like ELPP-3238, we probably should let users still authenticate rather than deny access. It means their profile will remain empty/incomplete.

### DIFF
--- a/profiles/api/oauth2.py
+++ b/profiles/api/oauth2.py
@@ -155,7 +155,11 @@ def create_blueprint(orcid: Dict[str, str], clients: Clients, profiles: Profiles
         try:
             orcid_record = orcid_client.get_record(profile.orcid, orcid_token.access_token)
             update_profile_from_orcid_record(profile, orcid_record)
-        except Exception as exception:
+        except RequestException as exception:
+            LOGGER.exception(exception)
+        except (LookupError, TypeError, ValueError) as exception:
+            # We appear to be misunderstanding the ORCID data structure, but let's not block the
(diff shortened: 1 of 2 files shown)


profiles/api/oauth2.py
```diff
@@ -155,7 +155,7 @@ def _token() -> Response:
         try:
             orcid_record = orcid_client.get_record(profile.orcid, orcid_token.access_token)
             update_profile_from_orcid_record(profile, orcid_record)
-        except RequestException as exception:
+        except Exception as exception:
```
giorgiosironi: suspicious of catching all exceptions e.g. an `NameError` from an undefined variable. Can we be more precise?

thewilkybarkid: Could be a variety of things, the overarching error is that the JSON structure isn't what we were expecting.

thewilkybarkid: `LookupError, TypeError, ValueError`?

seanwiseman: If you want to handle them all in the same way then grouping is fine:
`except (LookupError, TypeError, ValueError) as exception:`
Else you can chain them and handle each differently if you want to add more context:
```
try:
     .....
except LookupError as exception:
    # handle error
except TypeError as exception:
    # handle error

etc..
```

giorgiosironi: :+1: whitelisting over bug hiding

PR #48, opened 2017-10-31
ELPP-3224 Add cache headers

thewilkybarkid:


thewilkybarkid:
I tried to add a decorator, but couldn't get it to work (not able to work out what the problem is).

profiles/api/api.py
```diff
@@ -46,7 +46,13 @@ def _list() -> Response:
 
         response = make_response(json.dumps({'total': total, 'items': profile_list},
                                             default=normalize))
+        response.headers['Cache-Control'] = 'max-age=300, public, stale-if-error=86400,' \
```
giorgiosironi: module-level constant could help?
Or perhaps just a module-level function to apply standard headers to `response`?

test/test_api.py
```diff
@@ -156,14 +174,38 @@ def test_get_profile(test_client: FlaskClient) -> None:
...
 
     assert validate_json(data, schema_name='profile.v1') is True
     assert data['id'] == 'a1b2c3d4'
 
 
+def test_get_profile_revalidation(test_client: FlaskClient) -> None:
+    profile = Profile('a1b2c3d4', Name('Foo Bar'), '0000-0002-1825-0097')
+
+    db.session.add(profile)
+    db.session.commit()
```
giorgiosironi: would extract the 3 lines into something like `_add_profile('a1b2c3d4')`

profiles/api/api.py
seanwiseman: I have added two decorators and pushed a commit to this branch. Take a look and see if they fit what we need.

They are currently in the modules they are used in, these should probably be moved into a decorators module of some sort. There may be additional things you what to add to them, or additional decorators you what to make, if so they can be stacked.

test/test_api.py
seanwiseman: I have added some `@cache_control_headers` decorators and pushed them to this branch. Let me know if they fit what we need.

They will probably need renaming and moving into a decorators module of some sort, but its a start.

profiles/api/api.py
```diff
@@ -16,10 +16,20 @@
...
+        return response
+    return wrapper
+
+
 def create_blueprint(profiles: Profiles) -> Blueprint:
     blueprint = Blueprint('api', __name__)
 
-    @blueprint.route('/profiles')
+    @blueprint.route('/profiles', endpoint='_list')
+    @cache_control_headers
```
thewilkybarkid: That's what I was aiming for, but with an optional argument to say don't cache (for the OAuth routes).

Following advice on t'interweb was trying things like wrapt, but could never get it to work. Looks like that was never needed...

(Or was adding `endpoint` the fix?)

seanwiseman: The decorator effectively scraps the function name in a way, so the `@blueprint.route` was complaining as it thought you (we) was trying to register the same function for multiple routes. Defining the name in the `endpoint` arg sorted that, but I have just cleaned it up further by using `@wraps` on the internal wrapper function, this in turn keeps the name etc safe, removing the need for the `endpoint` arg.

You can add args to the decorators but it needs amending to handle that. Show me a sudo code sample and I can do it if you like.

PR #49, opened 2017-10-31
Type fixes

thewilkybarkid:
Fixing a few warnings from my IDE.

### DIFF
--- a/profiles/serializer/normalizer.py
+++ b/profiles/serializer/normalizer.py
@@ -5,13 +5,13 @@ from profiles.models import Affiliation, EmailAddress, Name, Profile
 
 
 @singledispatch
-def normalize(value: Any) -> str:
-    return str(value)
+def normalize(value: Any) -> Any:
+    return value
 
 
 @singledispatch
-def normalize_snippet(value: Any) -> str:
-    return str(value)
+def normalize_snippet(value: Any) -> Any:
+    return value
 
 
 @normalize.register(Profile)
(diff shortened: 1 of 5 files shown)


profiles/serializer/normalizer.py
```diff
@@ -5,13 +5,13 @@
 
 
 @singledispatch
-def normalize(value: Any) -> str:
-    return str(value)
+def normalize(value: Any) -> Any:
```
thewilkybarkid: No need to turn integers into strings. (This should really be string, numeric, dictionary, list, boolean or none.)

test/test_oauth2.py
```diff
@@ -211,14 +211,14 @@ def test_it_exchanges(generate_random_string: MagicMock, test_client: FlaskClien
...
     with requests_mock.Mocker() as mocker:
-        def token_text(request: Request):
+        def token_text(request: PreparedRequest):
             return urlencode({
                 'client_id': 'server_client_id',
                 'client_secret': 'server_client_secret',
                 'redirect_uri': 'http://localhost/oauth2/check',
                 'grant_type': 'authorization_code',
                 'code': '1234',
-            }) == request.text
```
thewilkybarkid: `text` came from `_RequestObjectProxy`, which is the actual type passed. It doesn't exist on `Request`.

profiles/serializer/normalizer.py
thewilkybarkid: Could introduce https://github.com/python/typing/issues/182?

seanwiseman: I think its clear enough as is, these will only be used via the `json` module calls anyway so the expected values should be clear enough. If MyPy aren't pushing it any further at the moment I think we are fine. Interesting topic though.

PR #51, opened 2017-10-31
ELPP-3195 Match existing profiles by email address

thewilkybarkid:


profiles/api/oauth2.py
```diff
@@ -133,22 +133,36 @@ def _token() -> Response:
...
         profile = _find_and_update_profile(json_data)
         json_data['id'] = profile.id
-        orcid_token = _find_and_update_access_token(json_data)
-
-        _update_profile(profile, orcid_token)
+        _find_and_update_access_token(json_data)
 
         return make_response(jsonify(json_data), response.status_code)
 
     def _find_and_update_profile(token_data: dict) -> Profile:
```
thewilkybarkid: This is getting a bit messy. Might be possible to refactor a bit. It will become easier when web hooks are in place and existing users don't have to be updated.

seanwiseman: Yes could perhaps move some of the model fetching and filtering into custom `@classmethods` on the model or repository, at least then there is the option of reusing the functionality if required in other views in the future.

profiles/api/oauth2.py
```diff
@@ -133,22 +133,36 @@ def _token() -> Response:
...
-            profile = Profile(profiles.next_id(), Name(token_data['name']), token_data['orcid'])
-            profiles.add(profile)
+            orcid_record = _fetch_orcid_record(token_data['orcid'], token_data['access_token'])
+
+            profile = None
+
+            if orcid_record:
+                for email_address in extract_email_addresses(orcid_record):
+                    try:
+                        profile = profiles.get_by_email_address(email_address['email'])
```
thewilkybarkid: Could have a `find_by_email_address` and allow multiple values?

profiles/api/oauth2.py
thewilkybarkid: I've refactored slightly, should we leave it for the moment given this area will have to change shortly anyway?

seanwiseman: Yer that sounds fine.

PR #52, opened 2017-11-01
ELPP-3205 Emit notifications when Profiles change

seanwiseman:
I have included functionality to configure, initialise and publish via `elife_bus_sdk`.

I need assistance with making sure that the db commit(s) are being executed in the right place, due to `SQLALCHEMY_COMMIT_ON_TEARDOWN`now being set to `False` to allow confirmation that a transaction has been successful before publishing an event to the bus. All the tests are passing after I mocked out the publisher, but there might be things that aren't being captured by the current tests now that `SQLALCHEMY_COMMIT_ON_TEARDOWN` is off.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,5 +1,5 @@
 [MESSAGE CONTROL]
-disable = abstract-method,missing-docstring,redefined-outer-name,too-few-public-methods, too-many-arguments
+disable = abstract-method,missing-docstring,redefined-outer-name,too-few-public-methods
 
 [TYPECHECK]
 ignored-classes = responses,scoped_session,SQLAlchemy
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,5 +1,5 @@
 [MESSAGE CONTROL]
-disable = abstract-method,missing-docstring,redefined-outer-name,too-few-public-methods
+disable = abstract-method,missing-docstring,redefined-outer-name,too-few-public-methods, too-many-arguments
 
 [TYPECHECK]
 ignored-classes = responses,scoped_session,SQLAlchemy
--- a/app.cfg.dist
+++ b/app.cfg.dist
@@ -17,4 +17,3 @@ client_secret = client_secret
 region = us-east-1
 subscriber = 512686554592
 name = bus-profiles
-env = dev
--- a/app.cfg.dist
+++ b/app.cfg.dist
@@ -12,3 +12,9 @@ authorize_uri = https://sandbox.orcid.org/oauth/authorize
 token_uri = https://sandbox.orcid.org/oauth/token
 client_id = client_id
 client_secret = client_secret
+
+[bus]
+region=us-east-1
+subscriber=1234567890
+name=profile
+env=dev
--- a/app.cfg.dist
+++ b/app.cfg.dist
@@ -14,7 +14,7 @@ client_id = client_id
 client_secret = client_secret
 
 [bus]
-region=us-east-1
-subscriber=512686554592
-name=bus-profiles
-env=dev
+region = us-east-1
+subscriber = 512686554592
+name = bus-profiles
+env = dev
--- a/app.cfg.dist
+++ b/app.cfg.dist
@@ -15,6 +15,6 @@ client_secret = client_secret
 
 [bus]
 region=us-east-1
-subscriber=1234567890
-name=profile
+subscriber=512686554592
+name=bus-profiles
 env=dev
--- a/manage.py
+++ b/manage.py
@@ -2,7 +2,6 @@ import configparser
 import logging
 import os
 
-from elife_bus_sdk import get_publisher
 from flask_migrate import MigrateCommand
 from flask_script import Manager, Server, Shell
 import yaml
@@ -22,14 +21,6 @@ CLIENTS = Clients(*[Client(name, **CLIENTS_DATA[name]) for name in CLIENTS_DATA]
 
 CONFIG = create_app_config(CONFIG_FILE)
 configure_logging(CONFIG.name, getattr(logging, CONFIG.logging['level']), CONFIG.logging['path'])
-
-PUBLISHER = get_publisher(pub_name='profiles',
-                          config={
-                              'region': CONFIG.bus['region'],
-                              'subscirber': CONFIG.bus['subscriber'],
-                              'name': CONFIG.bus['name'],
-                              'env': CONFIG.bus['env']
-                          })
 
 APP = create_app(CONFIG, CLIENTS)
 MANAGER = Manager(APP)
--- a/manage.py
+++ b/manage.py
@@ -2,6 +2,7 @@ import configparser
 import logging
 import os
 
+from elife_bus_sdk import get_publisher
 from flask_migrate import MigrateCommand
 from flask_script import Manager, Server, Shell
 import yaml
@@ -21,6 +22,15 @@ CLIENTS = Clients(*[Client(name, **CLIENTS_DATA[name]) for name in CLIENTS_DATA]
 
 CONFIG = create_app_config(CONFIG_FILE)
 configure_logging(CONFIG.name, getattr(logging, CONFIG.logging['level']), CONFIG.logging['path'])
+
+PUBLISHER = get_publisher(pub_name='profiles',
+                          config={
+                              'region': CONFIG.bus['region'],
+                              'subscirber': CONFIG.bus['subscriber'],
+                              'name': CONFIG.bus['name'],
+                              'env': CONFIG.bus['env']
+                          })
+
 APP = create_app(CONFIG, CLIENTS)
 MANAGER = Manager(APP)
 
--- a/profiles/api/oauth2.py
+++ b/profiles/api/oauth2.py
@@ -4,6 +4,8 @@ from json import JSONDecodeError, dumps
 from typing import Dict
 from urllib.parse import urlencode
 
+from elife_bus_sdk import get_publisher
+from elife_bus_sdk.events import ProfileEvent
 from flask import Blueprint, json, jsonify, make_response, redirect, request, url_for
 import requests
 from requests import RequestException
@@ -12,6 +14,7 @@ from werkzeug.wrappers import Response
 
 from profiles.clients import Clients
 from profiles.commands import update_profile_from_orcid_record
+from profiles.database import db
 from profiles.exceptions import ClientInvalidRequest, ClientInvalidScope, \
     ClientUnsupportedResourceType, InvalidClient, InvalidGrant, InvalidRequest, \
     OrcidTokenNotFound, ProfileNotFound, UnsupportedGrantType
@@ -21,6 +24,7 @@ from profiles.repositories import OrcidTokens, Profiles
 from profiles.utilities import expires_at, remove_none_values
 
 LOGGER = logging.getLogger(__name__)
+PUBLISHER = get_publisher('profiles')
 
 
 def create_blueprint(orcid: Dict[str, str], clients: Clients, profiles: Profiles,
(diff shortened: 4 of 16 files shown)


app.cfg.dist
```diff
@@ -12,3 +12,9 @@ authorize_uri = https://sandbox.orcid.org/oauth/authorize
 token_uri = https://sandbox.orcid.org/oauth/token
 client_id = client_id
 client_secret = client_secret
+
+[bus]
```
giorgiosironi: https://github.com/elifesciences/profiles-formula/blob/master/salt/profiles/config/srv-profiles-app.cfg needs updating, using some pillar values; `{{ pillar.elife.env}}` is good for `env`, `name` can be hardcoded, the other two need configuring.
These are the existing topics:
```
        {
            "TopicArn": "arn:aws:sns:us-east-1:512686554592:bus-profiles--ci"
        }, 
        {
            "TopicArn": "arn:aws:sns:us-east-1:512686554592:bus-profiles--continuumtest"
        }, 
        {
            "TopicArn": "arn:aws:sns:us-east-1:512686554592:bus-profiles--end2end"
        }, 
        {
            "TopicArn": "arn:aws:sns:us-east-1:512686554592:bus-profiles--prod"
        }, 
```

manage.py
```diff
@@ -21,6 +22,15 @@
 
 CONFIG = create_app_config(CONFIG_FILE)
 configure_logging(CONFIG.name, getattr(logging, CONFIG.logging['level']), CONFIG.logging['path'])
+
+PUBLISHER = get_publisher(pub_name='profiles',
+                          config={
+                              'region': CONFIG.bus['region'],
+                              'subscirber': CONFIG.bus['subscriber'],
```
giorgiosironi: typo?

profiles/api/oauth2.py
```diff
@@ -21,6 +24,7 @@
 from profiles.utilities import expires_at, remove_none_values
 
 LOGGER = logging.getLogger(__name__)
+PUBLISHER = get_publisher('profiles')
 
 
 def create_blueprint(orcid: Dict[str, str], clients: Clients, profiles: Profiles,
```
giorgiosironi: I think we should inject the publisher as an argument to `create_blueprint` to avoid global state

profiles/api/oauth2.py
```diff
@@ -137,6 +141,12 @@ def _token() -> Response:
 
         _update_profile(profile, orcid_token)
 
+        # manually commit session here?
+        db.session.commit()
```
giorgiosironi: perhaps this is good enough for now, it's a pain point to deal with sooner or later

profiles/config.py
```diff
@@ -9,7 +9,7 @@
 class Config(ABC):
     name = 'unknown'
     DEBUG = False
-    SQLALCHEMY_COMMIT_ON_TEARDOWN = True
+    SQLALCHEMY_COMMIT_ON_TEARDOWN = False
```
giorgiosironi: However if this is turned to `False` there are probably many places where we need to commit the transaction manually.
The pattern is the [Unit Of Work](https://martinfowler.com/eaaCatalog/unitOfWork.html), SQLAlchemy implements it with `db.session`, but perhaps we can automate the calls to that with some decorator (magic word) to put around the routes?

manage.py
seanwiseman: 👍🏻

profiles/api/oauth2.py
seanwiseman: The `publisher` is now being injected as an argument into the `create_blueprint` function.

app.cfg.dist
seanwiseman: I have opened PRs for both Profiles-formula and Builder-private including the new values.

profiles/factory.py
```diff
@@ -24,11 +25,14 @@ def create_app(config: Config, clients: Clients) -> Flask:
     orcid_client = OrcidClient(config.orcid['api_uri'])
     orcid_tokens = SQLAlchemyOrcidTokens(db)
     profiles = SQLAlchemyProfiles(db)
+
+    publisher = get_publisher(pub_name='profiles', config={**config.bus})
```
thewilkybarkid: `**config.bus` is a bit magic, could be explicit and remove `env` (since it's duplicating `profiles.environment`).

.pylintrc
```diff
@@ -1,5 +1,5 @@
 [MESSAGE CONTROL]
-disable = abstract-method,missing-docstring,redefined-outer-name,too-few-public-methods
+disable = abstract-method,missing-docstring,redefined-outer-name,too-few-public-methods, too-many-arguments
```
thewilkybarkid: I guess that's caused by `oauth2.create_blueprint()`? Is there are better way to do DI?

seanwiseman: Yes it was. You could wrap some of them up but the `create_blueprint` is creating the blueprint with x3 view functions and x3 utility functions so its having to accept args in turn for them all. If it was a 'normal' non factory function I would say reduce it, but I don't think its unacceptable to have 6 arguments for a function doing this much.

That said I normally follow fat models, thin views, so any business logic that deals with finding or updating a model stays in the model class (or possibly the `repository` in this case). That would leave the views a lot thinner and therefore probably require less arguments to initialise.

profiles/factory.py
seanwiseman: It was left to unpack due to the fact you can pass in additional arguments such as overriding `endpoint_url` and what not for the boto resource. (previous incarnations of this functionality included this so I did not want to get rid of it) 
```
# SNSPublisher
def __init__(self, region: str = '', subscriber: str = '', name: str = '', env: str = '', **overrides):
```
Though in `Profiles` at least its probably not going to need anything other than the values to build the Arn, so I don't mind being explicit. Will change. Will also using existing `profiles.env` value.

seanwiseman:
I have now moved the send event logic into a function that will be executed each time model(s) are inserted, updated or removed. This is done by hooking into SQLAlchemy signals so I have been able to remove any code from the views as its not needed there, and have reverted `SQLALCHEMY_COMMIT_ON_TEARDOWN` back to `True` as (for this case at least) there is not a need for me to manually commit the session.

.pylintrc
thewilkybarkid: Can be reverted now? (We might hit it later though.)

profiles/events.py
```diff
@@ -0,0 +1,27 @@
...
+from profiles.models import Profile
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+def send_update_events(publisher: EventPublisher) -> Callable[..., None]:
+    def wrapper(sender: Flask, changes: tuple) -> None:  # pylint:disable=unused-argument
+        for instance, operation in changes:
+            if isinstance(instance, Profile):
```
thewilkybarkid: Does this cater for something like just an `Affiliation` change? We could do with tests here.

.pylintrc
seanwiseman: Yes that can go now. Will remove.

profiles/factory.py
```diff
@@ -39,4 +47,6 @@ def create_app(config: Config, clients: Clients) -> Flask:
     app.register_error_handler(ClientError, errors.client_error_handler)
     app.register_error_handler(OAuth2Error, errors.oauth2_error_handler)
 
+    models_committed.connect(send_update_events(publisher=publisher), app)
```
thewilkybarkid: That's nice and easy to use.

profiles/events.py
seanwiseman: Currently I only act if the instance being saved is of type `Profile` though all model changes will pass through here so can be handled. Shall I add support for `Affiliation` and some tests surrounding it?

thewilkybarkid: It needs to be if _any_ part of a `Profile` changes (so `EmailAddress` too; what happens with composite types?)

seanwiseman: OK so `Profile`, `Affiliation` and `EmailAddress` will need to have slightly different handling but the composite type updates effectively count as field updates on the parent model they are attached to. So an `Address` thats changed and saved will trigger the associated `Affiliation` to be updated. I will add them and run the additional tests.

seanwiseman: OK so we now have support for updates of all models excluding `OrcidToken`.

I had refactor the event publisher function to keep a list of ids for a single `session.commit` and then filter out duplicates so you don't receive multiple messages for single events, for example:

- Updating a `Profile` directly would send one change event` [(<Profile test>, 'update')]` this is fine, one event is sent to SNS.

- Creating and adding an `Affiliation` would send two change events `[(<Affiliation 1>, 'insert'), (<Profile test>, 'update')]` in this instance only one event is sent to SNS because I filter out duplicate related profile ids.

- Updating an `Address` (or any composite types) would only send one change event (linked to the parent model) `[(<Affiliation 1>), 'update']` this is fine and one event will be sent to SNS.

In short, any sort of change or collection of changes to a single `Profile` will send only one event to SNS (including composite types).

profiles/events.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def send_update_events(publisher: EventPublisher) -> Callable[..., None]:
+    def wrapper(sender: Flask, changes: tuple) -> None:  # pylint:disable=unused-argument
+        ids = []
+
+        for instance, operation in changes:  # pylint:disable=unused-variable
+            if isinstance(instance, Profile):
+                ids.append(instance.id)
+            if isinstance(instance, (Affiliation, EmailAddress)):
```
giorgiosironi: It's a bit worrying that we would have to traverse the graphs from any modified object to go back to what is the `profile` being modified, but it's good for now :+1: 
It comes from the fact that the ORM is fine-grained for the single entities, but we are after a model where `1 transaction` == `1 Profile Aggregate` == `1 Profile object and any of the Value Object it contains`

seanwiseman: I agree. If it was not for composite types being used we could just look for `Profile` model changes as any direct change to any of its Foreign keys is picked up, its just that the Foreign key's composite types do not get classed as a `Profile` model update so you have to account of the other `db.models`.

But yes its fine for now.

thewilkybarkid: Composites are the problem? I'm not entirely happy with how they work, so we could just replace them with plain value objects. (Though this actually sounds like an upstream bug.)

seanwiseman: They are not a problem, they are just the current reason for having to check per model type. 

I am used to them being plain value objects, but there will still be an element of checking based on model types even if you do that. 

If say `Address` was a `db.Model` and had a foreign key relationship to an `Affiliation` and you updated the `Address` object, you would still only get updates up to the `Affiliation` (would not bubble up to a `Profile` update), that's why we have to pull the profile id from the `Affiliation` object in that case.

For this use case I think its fine, its only run on a `session.commit` and will only send one event message per profile update.

thewilkybarkid: Ah ok, I misunderstood.

thewilkybarkid:
Looks good, I think we just need two more tests: one if a profile gets deleted, and an integration test (ie `test_oauth2`).

seanwiseman:
I have added a test for profile deletion, I also made sure that the object information is still passed to the event (so we still have an `id` even if its just been deleted) and it is.

As for an integration test, really its just testing that SQLAlchemy actually delivers model event information when it says it should, its not our code per say as we already test how it handles these events when it receives them. Any how, I am getting into mock spaghetti at the moment trying to track the calls to the 'actual' callback function that was passed to the listener in the app in `create_app`. 
Will keep trying.

giorgiosironi:
For what it's worth we will also test events integration in end2end, but there is no one listening yet where we could make the assertion.

seanwiseman:
After further investigation its seems that event signals (handled by the `blinker` library) are not actually triggered when using app and db test fixtures. I even tried `SQLAlchemy` base events and they don't seem to fire in test cases either. When being used outside of the test environment they fire fine, its something to with how the fixtures are setup and torn down.

I have added a test to prove we have a handler registered on app init and then other tests to show how these events are handled, I just can't mock the 'sending' of the signals using the test db fixture at present. I think the end to end tests will cover what we need.

giorgiosironi:
They will, as soon as we have a client for these events to check.
An intermediate level is also to perform integration tests against the real nginx+uwsgi service running on `http://localhost:80`. Not sure we need such a layer here, but it's an option.

seanwiseman:
OK understood.

PR #53, opened 2017-11-15
ELPP-3197 Add ORCID webhook

thewilkybarkid:


profiles/events.py
```diff
@@ -1,16 +1,59 @@
...
+
+        data = {
+            'client_id': orcid['client_id'],
+            'client_secret': orcid['client_secret'],
+            'scope': '/webhook',
+            'grant_type': 'client_credentials',
+        }
+
+        response = requests.post(url=orcid['token_uri'], data=data,
+                                 headers={'Accept': 'application/json'})
```
thewilkybarkid: Might be better to cache the results of this, we (should) only need to fetch one once.

profiles/api/webhook.py
```diff
@@ -0,0 +1,29 @@
...
+
+    @blueprint.route('/orcid-webhook/<orcid>', methods=['POST'])
+    def _update(orcid: str) -> Response:
+        try:
+            profile = profiles.get_by_orcid(orcid)
+        except ProfileNotFound as exception:
+            raise NotFound(str(exception)) from exception
+
+        access_token = orcid_tokens.get(profile.orcid)
+        orcid_record = orcid_client.get_record(orcid, access_token.access_token)
```
thewilkybarkid: Need to cater for the access token being rejected.

seanwiseman: What status code would you want returned in the event of a token rejection?

thewilkybarkid: It's what ORCID would reject us with. Guessing it's a 403?

profiles/events.py
seanwiseman: I agree. Probably best to have this as a stand alone task as it will effectively be setting up a caching solution that any part of the app could use. It may need more thought as to what other caching requirements we anticipate.

profiles/api/webhook.py
```diff
@@ -20,7 +21,11 @@ def _update(orcid: str) -> Response:
             raise NotFound(str(exception)) from exception
 
         access_token = orcid_tokens.get(profile.orcid)
-        orcid_record = orcid_client.get_record(orcid, access_token.access_token)
+
+        try:
+            orcid_record = orcid_client.get_record(orcid, access_token.access_token)
+        except RequestException as exception:
+            raise Forbidden(str(exception)) from exception
```
thewilkybarkid: Here we should delete the stored access token and then make a second request to update the profile from their public data.

thewilkybarkid:
Tests need fixing, looks like the listener is trying to register webhooks in various places.

profiles/orcid.py
```diff
@@ -1,20 +1,52 @@
...
 
-import requests
-from requests import RequestException, Response
+from requests import RequestException, Response, post, request
 
 API_VERSION = 'v2.0'
 LOGGER = logging.getLogger(__name__)
 
+PRIVATE_SCOPE = '/webhook'
+PUBLIC_SCOPE = '/read-public'
```
thewilkybarkid: These are very confusing, they're not 'public' and 'private'.

profiles/repositories.py
```diff
@@ -66,6 +70,14 @@ def get(self, orcid: str) -> OrcidToken:
     def clear(self) -> None:
         self.db.session.query(OrcidToken).delete()
 
+    def remove(self, orcid: str) -> None:
+        try:
+            orcid_token = self.db.session.query(OrcidToken).filter_by(orcid=orcid).one()
+            self.db.session.delete(orcid_token)
+        except NoResultFound as exception:
+            raise OrcidTokenNotFound('ORCID token for the ORCID {} not found'.format(orcid)) \
```
thewilkybarkid: Not sure there's a need to throw an exception here? If you try and remove something that doesn't exist, 🤷‍♂️ .

profiles/orcid.py
seanwiseman: OK, do you have a suggestion? something like:

`WEBHOOK_SCOPE`
`READ_PUBLIC_SCOPE`

profiles/repositories.py
seanwiseman: OK change to `pass` or log the exception?

profiles/orcid.py
```diff
@@ -16,37 +16,21 @@
...
-    def __init__(self, api_uri: str, token_uri: str, client_id: str, client_secret: str) -> None:
+    # pylint:disable=too-many-arguments
+    def __init__(self, api_uri: str, token_uri: str, client_id: str, client_secret: str,
+                 read_public_access_token: str, webhook_access_token: str) -> None:
         self.api_uri = api_uri
         self.token_uri = token_uri
         self.client_id = client_id
         self.client_secret = client_secret
+        self.read_public_access_token = read_public_access_token
+        self.webhook_access_token = webhook_access_token
```
thewilkybarkid: I'd keep these out of the `OrcidClient` now.

seanwiseman: We are already passing an `OrcidClient` instance to the `webhook_maintainer` and the `_update` view function, if removed, we would then have to also provide the callers access to the `config` somehow. The `OrcidClient` is already being passed the rest of the values for orcid from the config so it seemed to fit to pass it these and provide the same API to the callers `orcid_client.get_access_token()`.

thewilkybarkid: We already do that elsewhere (eg `api/oauth2.py` has `orcid: Dict[str, str]`; `OrcidClient` is meant to be like a 3rd-party library so best not to require these two upfront - it's just not be extracted as its incomplete).

seanwiseman: OK understood. I have now refactored them out of the `OrcidClient`.

profiles/config.py
```diff
@@ -23,24 +23,29 @@ def __init__(self, orcid: Dict[str, str], db: str, logging: Dict[str, str],
...
 
 class CiConfig(DevConfig):
     TESTING = True
+    SERVER_NAME = 'ci--profiles.elifesciences.org'
     name = 'ci'
 
 
 class ProdConfig(Config):
     name = 'prod'
+    SERVER_NAME = 'prod--profiles.elifesciences.org'
```
thewilkybarkid: This should probably be configuration? /cc @giorgiosironi

thewilkybarkid: Might want to set `PREFERRED_URL_SCHEME` at the same time? (Default's `http`, but we use `https` everywhere outside of Vagrant.)

profiles/events.py
```diff
@@ -1,11 +1,5 @@
 import logging
-from typing import (
-    Any,
-    Callable,
-    Dict,
-    List,
-    Tuple,
-)
+from typing import Any, Callable, Dict, List, Tuple
```
seanwiseman: Fine to revert this, but I consciously made the change as its seen to be idiomatic to wrap 3 or more imports in a single line into a tuple, e.g. 

https://github.com/django/django/blob/master/django/http/multipartparser.py

It is not enforced by the linters (as far as I know) but is common place in major projects/libraries.

thewilkybarkid: Ah ok, was just aiming for a smaller diff (and it hasn't been done elsewhere).

seanwiseman: OK granted, there are a few files this could be applied to. Just leave it for now then.

profiles/api/oauth2.py
```diff
@@ -134,30 +134,31 @@ def _token() -> Response:
...
 
-        if token_data['name']:
-            profile.name = Name(token_data['name'])
+    def _create_profile(token_data: dict, orcid_record: dict) -> Profile:
+        if not token_data['name']:
+            raise InvalidRequest('No name visible')
+        profile = Profile(profiles.next_id(), Name(token_data['name']), token_data['orcid'])
+        profiles.add(profile)
+
+        profile.name = Name(token_data['name'])
```
giorgiosironi: the behavior change seems to never update the name here, assigning it only on creation; in the case of a profile without an ORCID being found through email, it could keep the lower quality (e.g. Disqus) name around?

thewilkybarkid: Yep, need to add an `_update_profile()` call there.

profiles/api/webhook.py
```diff
@@ -0,0 +1,43 @@
...
+        try:
+            orcid_record = orcid_client.get_record(orcid, access_token)
+        except RequestException as exception:
+            if exception.response.status_code == 403 and not access_token == \
+                    orcid_config.get('read_public_access_token'):
+                orcid_tokens.remove(profile.orcid)
+
+            raise exception
+
+        update_profile_from_orcid_record(profile, orcid_record)
```
giorgiosironi: flow is clear

profiles/api/webhook.py
```diff
@@ -0,0 +1,43 @@
...
+from profiles.exceptions import OrcidTokenNotFound, ProfileNotFound
+from profiles.orcid import OrcidClient
+from profiles.repositories import OrcidTokens, Profiles
+
+
+def create_blueprint(profiles: Profiles, orcid_config: Dict[str, str],
+                     orcid_client: OrcidClient, orcid_tokens: OrcidTokens) -> Blueprint:
+    blueprint = Blueprint('webhook', __name__)
+
+    @blueprint.route('/orcid-webhook/<orcid>', methods=['POST'])
```
giorgiosironi: any authentication on this route? It looks like it could be used to flood ORCID with requests coming from us

test/test_api.py
```diff
@@ -289,7 +299,8 @@ def test_it_does_not_return_restricted_affiliations(test_client: FlaskClient, ye
     profile.add_affiliation(affiliation)
     profile.add_affiliation(affiliation2)
 
-    db.session.commit()
+    with patch('profiles.orcid.request'):
```
giorgiosironi: smells of a duplicate concept, can these two lines be extracted to cater for whatever else we want to mock?

profiles/api/webhook.py
thewilkybarkid: Not at the moment, _think_ only possibility is basic auth or restricting access to ORCID's IPs.

test/test_webhook.py
```diff
@@ -0,0 +1,88 @@
...
+def test_it_returns_403_if_an_access_token_is_rejected(profile: Profile, orcid_token: OrcidToken,
+                                                       test_client: FlaskClient) -> None:
+    db.session.add(profile)
+    db.session.add(orcid_token)
+
+    with patch('profiles.orcid.request'):
+        db.session.commit()
+
+    with requests_mock.Mocker() as mocker:
+        mocker.get('http://www.example.com/api/v2.1/0001-0002-1825-0097/record', exc=Forbidden)
```
giorgiosironi: not sure who throws `Forbidden`? It's not very clear what is the difference between this test case and the next, since I'd assume (perhaps wrongly) they both deal with a `403 Forbidden` from ORCID

test/test_api.py
seanwiseman: It's preventing HTTP calls from the `OrcidClient` that is used through the registered `webhook_maintainer` in the `create_app` call. This was the cleanest way to prevent the calls (just mocking it when `commit` is called). I tried mocking out the passed in `OrcidClient` in the `create_app` call (this needs to be done in the `conftest` as the `app` has a session scope) but it got messy and felt like it was muddying the whole `app` creation which had knock-on effects on other tests. Also sometimes you might not want to mock this and intercept or inspect the HTTP call.

Open to suggestions.

giorgiosironi: Just meant to extract:
```
def _commit():
    with patch('profiles.orcid.request'):
         db.session.commit()
```
in this file

test/test_webhook.py
seanwiseman: The first concentrates on just if a 403 is thrown, the second test focuses on if the following conditional was hit:
`if exception.response.status_code == 403 and not access_token == orcid_config.get('read_public_access_token'):`

test/test_api.py
seanwiseman: Ah ok, well I didn't want to start pulling snippets into functions (in local files) as we have not done this throughout the test suite, everything has been structured as fixtures in `conftest` and injected.

That said I agree with the suggestion, I have made the change. If we start requiring this in other test files it may need to be moved/refactored.

profiles/api/webhook.py
thewilkybarkid: How would you rather tackle this (if at all)?

giorgiosironi: The only straightforward option seems to be to add a secret query string parameter, as that can be passed as part of the webhook URL. But there is no easy way to update it, would have to re-register all webhooks if it has to change.

giorgiosironi: Slightly more elaborate: add a signature to the webhook URL?

giorgiosironi: (wouldn't do it in this PR)

test/test_webhook.py
```diff
@@ -59,28 +57,44 @@ def test_it_returns_403_if_an_access_token_is_rejected(profile: Profile, orcid_t
         db.session.commit()
 
     with requests_mock.Mocker() as mocker:
-        mocker.get('http://www.example.com/api/v2.1/0001-0002-1825-0097/record', exc=Forbidden)
```
thewilkybarkid: `Forbidden` was a `RequestException` anyhow.

profiles/api/webhook.py
thewilkybarkid: @seanwiseman It's to stop the endpoint being publicly accessible, as it triggers work both locally and for ORCID.

Could be part of the path rather than query string (`/orcid-webhook/<secret>/<orcid>`).

If we do want to do this and not do it IP based, I suggest we _do_ add it now as updating would be a pain (`DELETE` then a `POST` for every profile). We should probably put a comment next to the value in builder-private saying that changing the value causes problems.

seanwiseman: Ah, I see, thank you.

giorgiosironi: Proof of concept:
```
import hmac 
import base64

# open("/dev/urandom","rb").read(32)
key = b'\xd2\r\x94\xde\xef\x96!l+[B\xf4\xa5p\xa2\x90\xe01c\xc5{\xe1K\x03\x02\x11X\r\xc0\xb7\xb2\xf7'

# original URL
url = 'https://prod--profiles.elifesciences.org/orcid/0000-0000-0000-0000'
# generate signature
h = hmac.new(key, url.encode('utf-8'), 'sha256')
digest = base64.b64encode(h.digest())
signed_url = '%s?signature=%s' % (url, digest)
print(digest)
print(signed_url)

# validating signature on an incoming request                                                                                              
received = base64.b64decode(digest) # from signature parameter
h = hmac.new(key, url.encode('utf-8'), 'sha256') # from rest of URL
print(hmac.compare_digest(h.digest(), received))
```

A signature is superior to a secret parameter here, as a URL dropped in Slack does not compromise all other URLs but only that one.

thewilkybarkid: Definitely superior (and very cautious; ORCID docs provide no recommendations for this). http://flask.pocoo.org/snippets/50/ seems to integrate nicely with `url_for()`, so doesn't make it harder to use.

profiles/factory.py
```diff
@@ -28,6 +29,8 @@ def create_app(config: Config, clients: Clients) -> Flask:
     orcid_tokens = SQLAlchemyOrcidTokens(db)
     profiles = SQLAlchemyProfiles(db)
 
+    uri_signer = URLSafeSerializer(config.orcid['webhook_key'])
```
giorgiosironi: it seems the default in our version is SHA-1, but `master` that uses SHA-512 does not have a release yet:
https://github.com/pallets/itsdangerous/blob/0.24/itsdangerous.py#L297
https://github.com/pallets/itsdangerous/blob/master/itsdangerous.py#L308
We should specify SHA-512 ourselves not to rely on a default that is going to change.

Also `key_derivation` should be `hmac`, not the Django default https://github.com/pallets/itsdangerous/blob/0.24/itsdangerous.py#L307

giorgiosironi:
Needs a formula addition though.

profiles/cli.py
```diff
@@ -18,3 +31,28 @@ def __init__(self, *repositories: CanBeCleared) -> None:
...
+    # let's see whether we get more commands
+    # and we keep flask_script in the long term
+    # pylint: disable=method-hidden
+    def run(self) -> None:
+        access_token = self.orcid.get('webhook_access_token')
+
+        for profile in self.profiles.list():
+            uri = url_for('webhook._update', payload=self.uri_signer.dumps(profile.orcid),
+                          _external=True)
+            self.orcid_client.set_webhook(profile.orcid, uri, access_token)
```
giorgiosironi: it would be useful to log/print something to understand how many profiles have been involved, how quickly it is progressing

thewilkybarkid: `OrcidClient.set_webhook()` already prints what its doing, but there's no count yet. Will add one.

thewilkybarkid:
Merge? 🤞

PR #55, opened 2017-11-22
Fix event sending

seanwiseman:


profiles/events.py
```diff
@@ -16,16 +16,23 @@ def send_update_events(publisher: EventPublisher) -> Callable[..., None]:
     def event_handler(sender: Flask, changes: tuple) -> None:  # pylint:disable=unused-argument
         ids = []
 
+        LOGGER.debug('Processing event(s)')
```
giorgiosironi: Speaking of the log level, this could easily be `info` as it should be left turned on in `profiles--prod` for a long time. We have no good log at the SNS level of what is going on, only some monitoring on the SQS queues that listen to the topic.

giorgiosironi: My rule of thumb is `if it doesn't fill the disk, then it's info`

giorgiosironi:
Fixed the build, was deploying a refactoring of the formula.

profiles/events.py
```diff
@@ -16,16 +16,23 @@ def send_update_events(publisher: EventPublisher) -> Callable[..., None]:
     def event_handler(sender: Flask, changes: tuple) -> None:  # pylint:disable=unused-argument
```
thewilkybarkid: `sender` type needs fixing.

profiles/events.py
```diff
@@ -16,16 +16,23 @@ def send_update_events(publisher: EventPublisher) -> Callable[..., None]:
...
         ids = []
 
+        LOGGER.info('Processing event(s)')
+
         for instance, operation in changes:  # pylint:disable=unused-variable
+            LOGGER.info('Event: %s - %s', instance, operation)
```
thewilkybarkid: Probably should be something like `'Found operation %s %s', operation, instance`?

profiles/events.py
```diff
@@ -16,16 +16,23 @@ def send_update_events(publisher: EventPublisher) -> Callable[..., None]:
...
             if isinstance(instance, Profile):
                 ids.append(instance.id)
             if isinstance(instance, (Affiliation, EmailAddress)):
                 ids.append(instance.profile.id)
 
         for profile_id in set(ids):
             try:
+                LOGGER.info('Sending event for profile id: %s', profile_id)
```
thewilkybarkid: Bit inconsistent with the failure message, should be `Sending event for Profile %s`?

profiles/factory.py
```diff
@@ -47,6 +47,6 @@ def create_app(config: Config, clients: Clients) -> Flask:
     app.register_error_handler(ClientError, errors.client_error_handler)
     app.register_error_handler(OAuth2Error, errors.oauth2_error_handler)
 
-    models_committed.connect(send_update_events(publisher=publisher), app, weak=False)
+    models_committed.connect(send_update_events(publisher=publisher), weak=False)
```
thewilkybarkid: Is it possible to test this?

profiles/events.py
seanwiseman: Ah, was corrected in https://github.com/elifesciences/profiles/pull/53 but not here. Now done.

profiles/events.py
seanwiseman: Changed.

profiles/events.py
seanwiseman: Changed.

profiles/events.py
thewilkybarkid: Was referring to `sender: Flask`, which is now `sender: Any`.

profiles/events.py
```diff
@@ -13,19 +14,26 @@
...
 
 def send_update_events(publisher: EventPublisher) -> Callable[..., None]:
-    def event_handler(sender: Flask, changes: tuple) -> None:  # pylint:disable=unused-argument
+    def event_handler(sender: Flask, changes: List[Tuple[db.Model, str]]) -> None:  # pylint:disable=unused-argument
         ids = []
 
+        LOGGER.info('Processing event(s)')
+
         for instance, operation in changes:  # pylint:disable=unused-variable
+            LOGGER.info('Found operation %s %s', instance, operation)
```
thewilkybarkid: I'd put `instance` and `operation` the other way around.

profiles/factory.py
giorgiosironi: If we had a route to POST to whose only behavior was to do an empty update of a `Profile` (bumping its updated date, if there is one?), we could do a smoke test or a manual test that would work on `dev` and `ci`.

Otherwise this will be tested in end2end, but we have no checks in automated test yet for that as there is no observable behavior depending on the event (that I have seen).

seanwiseman: I could take another look, but this was a sticking point last time. Events just don't seem to get fired in the same way when using test fixtures, when run using a dummy app (live not test suite) it works. Something to do with the setup/teardown + blinker. That's why we wanted some kind of end2end test(s) for this.

seanwiseman: I take that back. Replacing Signal.ANY was, in fact, the solution for everything. Can easily test this now. Will add additional tests.

Thanks @thewilkybarkid

profiles/events.py
```diff
@@ -14,13 +15,13 @@
 
 
 def send_update_events(publisher: EventPublisher) -> Callable[..., None]:
-    def event_handler(sender: Flask, changes: List[Tuple[db.Model, str]]) -> None:  # pylint:disable=unused-argument
+    def event_handler(sender: Signal.ANY, changes: List[Tuple[db.Model, str]]) -> None:  # pylint:disable=unused-argument
```
thewilkybarkid: `Typing.Any`?

profiles/events.py
```diff
@@ -13,19 +14,27 @@
 
 
 def send_update_events(publisher: EventPublisher) -> Callable[..., None]:
-    def event_handler(sender: Flask, changes: tuple) -> None:  # pylint:disable=unused-argument
+    # pylint:disable=unused-argument
+    def event_handler(sender: Signal.ANY, changes: List[Tuple[db.Model, str]]) -> None:
```
thewilkybarkid: `typing.Any`?

seanwiseman:
Overthinking database transaction event signals by Sean Wiseman....

PR #56, opened 2017-11-27
Update ORCID API to 2.1

thewilkybarkid:
Doesn't look like we need to make changes: https://orcid.org/blog/2017/11/16/announcing-api-21-orcid-ids-are-now-https


seanwiseman:
We could maybe add a `dummy_api_root_url` fixture to avoid maintenance in the future, even though it's only applicable to 9 statements or so at present:

```
conftest.py:

from profiles.orcid import API_VERSION

@fixture
def dummy_api_root_url():
    return '{root_url}/{version}'.format(root_url='**http://someurl.com**', version=API_VERSION)

thewilkybarkid:
Like the idea (I'd keep setting `API_VERSION` manually though); probably should be applied to all the URIs. But, since it's a quick search+replace that doesn't really take any longer, we can refactor later.

giorgiosironi:
Any change to https://github.com/elifesciences/orcid-dummy/blob/master/src/bootstrap.php#L50-L238 ? Probably no change.

thewilkybarkid:
Only change seems to be `http://sandbox.orcid.org/0000-0001-8778-8651` to `https://sandbox.orcid.org/0000-0001-8778-8651`.

giorgiosironi:
Depends on https://github.com/elifesciences/orcid-dummy/pull/6 to go through end2end.

PR #58, opened 2017-11-28
ELPP-3304 Add access levels to profiles

thewilkybarkid:


test/test_api.py
```diff
@@ -243,49 +226,20 @@ def test_does_not_contain_restricted_email_addresses(test_client: FlaskClient) -
 
     assert response.status_code == 200
     assert response.headers.get('Content-Type') == 'application/vnd.elife.profile+json;version=1'
-    assert validate_json(data, schema_name='profile.v1') is True
-    assert data['emailAddresses'] == ['1@example.com']
+    # assert validate_json(data, schema_name='profile.v1') is True
```
thewilkybarkid: Requires https://github.com/elifesciences/api-raml/pull/204 to be merged.

profiles/serializer/normalizer.py
```diff
@@ -17,34 +20,40 @@ def normalize_snippet(value: Any) -> Any:
 @normalize.register(Profile)
 def normalize_profile(profile: Profile) -> dict:
     data = normalize_profile_snippet(profile)
-    data['emailAddresses'] = [normalize(email) for email in profile.get_email_addresses()]
-    data['affiliations'] = [normalize(aff) for aff in profile.get_affiliations()]
+    data['emailAddresses'] = [normalize(email) for email in profile.get_email_addresses(True)]
+    data['affiliations'] = [normalize(aff) for aff in profile.get_affiliations(True, True)]
```
thewilkybarkid: We could get rid of the `include_restricted` option now as it's no longer needed?

giorgiosironi: Depends on how the design of the access control ends up filtering the values, if it does it on `data` than the option is not needed but if it wants to push it down to `profile`...

profiles/serializer/normalizer.py
```diff
@@ -17,34 +20,40 @@ def normalize_snippet(value: Any) -> Any:
...
-            "formatted": [
-                affiliation.address.city,
-                affiliation.address.region,
-                affiliation.address.country.name,
-            ],
-            "components": {
-                "locality": [affiliation.address.city],
-                "area": [affiliation.address.region],
-                "country": affiliation.address.country.name
+        'access': ACCESS_RESTRICTED if affiliation.restricted else ACCESS_PUBLIC,
```
giorgiosironi: being this *only* the schema change (and having no authentication yet), shouldn't we include *only* the public data, and separately added authentication to expose `restricted` values?
The authentication works with a combination of magic https://github.com/elifesciences/builder-private/blob/master/pillar/api-gateway.sls#L160 and more magic https://github.com/elifesciences/lax-formula/blob/master/salt/lax/config/etc-nginx-sitesavailable-lax.conf#L18-L25 so will need a review and genericization

profiles/serializer/normalizer.py
thewilkybarkid: Err, yeah, need to hide it if not authorised. Long day...

profiles/serializer/normalizer.py
thewilkybarkid: We'll do a follow-up to expose `restricted` values as nothing will use them (yet).

PR #59, opened 2017-11-29
Profile list API should only return snippets

thewilkybarkid:


test/test_api.py
```diff
@@ -60,6 +60,21 @@ def test_list_of_profiles(test_client: FlaskClient) -> None:
...
+    profile.add_email_address('foo@example.com')
+
+    db.session.add(profile)
+    db.session.commit()
+
+    response = test_client.get('/profiles')
+
+    data = json.loads(response.data.decode('UTF-8'))
+
+    assert validate_json(data, schema_name='profile-list.v1') is True
```
seanwiseman: Realised we are not using the actual response validator in the api-validator lib, we are still using the utility function. I can update this in another PR.

PR #60, opened 2017-11-29
Use JMESPath to pick data out of the ORCID API response

thewilkybarkid:
Seems unclear whether the ORCID responses should omit values or leave them as `null`, but they do the latter. JMESPath is probably a better way to traverse the uncertain structure (this doesn't do it _everywhere_).

### DIFF
--- a/profiles/commands.py
+++ b/profiles/commands.py
@@ -1,7 +1,7 @@
 from typing import List, Optional
 
-import jmespath
 from iso3166 import countries
+import jmespath
 
 from profiles.models import Address, Affiliation, Date, Name, Profile
 from profiles.orcid import VISIBILITY_PUBLIC
--- a/profiles/commands.py
+++ b/profiles/commands.py
@@ -23,17 +23,16 @@ def extract_email_addresses(orcid_record: dict, only_verified: bool = True) -> L
 
 
 def _update_name_from_orcid_record(profile: Profile, orcid_record: dict) -> None:
-    if 'name' in orcid_record.get('person', {}):
-        given_name = jmespath.search('person.name."given-names".value', orcid_record)
-        family_name = jmespath.search('person.name."family-name".value', orcid_record)
+    given_name = jmespath.search('person.name."given-names".value', orcid_record)
+    family_name = jmespath.search('person.name."family-name".value', orcid_record)
 
-        if given_name and family_name:
-            profile.name = Name('{} {}'.format(given_name, family_name),
-                                '{}, {}'.format(family_name, given_name))
-        elif given_name:
-            profile.name = Name(given_name, given_name)
-        elif family_name:
-            profile.name = Name(family_name, family_name)
+    if given_name and family_name:
+        profile.name = Name('{} {}'.format(given_name, family_name),
+                            '{}, {}'.format(family_name, given_name))
+    elif given_name:
+        profile.name = Name(given_name, given_name)
+    elif family_name:
+        profile.name = Name(family_name, family_name)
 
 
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -2,10 +2,15 @@ alembic==0.9.5
 -e git+https://github.com/elifesciences/bus-sdk-python.git@49a0b0ecc55f99bc55a61f3e9b613ea88e3ff5a5#egg=bus-sdk-python
 -e git+https://github.com/elifesciences/api-validator-python.git@2971a2146321d48cc88c9c9f36e8fd72740dbfee#egg=api-validator-python
 astroid==1.5.3
+attrs==17.3.0
 blinker==1.4
+boto3==1.4.7
+botocore==1.7.36
```
thewilkybarkid: No `pip freeze` recently? (Maybe CI should check?)

profiles/commands.py
```diff
@@ -13,7 +14,7 @@ def update_profile_from_orcid_record(profile: Profile, orcid_record: dict) -> No
 
 
 def extract_email_addresses(orcid_record: dict, only_verified: bool = True) -> List[dict]:
-    orcid_emails = orcid_record.get('person', {}).get('emails', {}).get('email', {})
+    orcid_emails = jmespath.search('person.emails.email[*]', orcid_record) or []
```
thewilkybarkid: Seems to return `None` rather than an empty list.

profiles/commands.py
```diff
@@ -23,8 +24,8 @@ def extract_email_addresses(orcid_record: dict, only_verified: bool = True) -> L
 
 def _update_name_from_orcid_record(profile: Profile, orcid_record: dict) -> None:
     if 'name' in orcid_record.get('person', {}):
-        given_name = orcid_record['person']['name'].get('given-names', {}).get('value')
-        family_name = orcid_record['person']['name'].get('family-name', {}).get('value')
+        given_name = jmespath.search('person.name."given-names".value', orcid_record)
+        family_name = jmespath.search('person.name."family-name".value', orcid_record)
```
giorgiosironi: needs 3 test cases:

- missing `given-names` and/or `family-name` fields
- fields are present, but with value `null`
- fields are present, but with value `{"value":""}`

requirements.txt
giorgiosironi: that `pip freeze`'s output is identical to this file?

thewilkybarkid:
I've left https://github.com/elifesciences/profiles/blob/develop/profiles/api/oauth2.py#L189-L192 in place, just in case.

giorgiosironi:
Outdated comment, but it's not:
```
needs 3 test cases:

    missing given-names and/or family-name fields
    fields are present, but with value null
    fields are present, but with value {"value":""}
```

requirements.txt
thewilkybarkid: Yeah, so it's always up to date.

thewilkybarkid:
I suspect a blank string would cause issues.

profiles/commands.py
seanwiseman: `family_name = jmespath.search('person.name."family-name".value', orcid_record) or []` returns `[]` when tested so it may be something to do with the `email[*]` part of the expression. Would expect the right side value to be assigned if `.search()` returned falsely.

thewilkybarkid: Reading the JMESPath spec think it will always be `None` if not set at all, so this will have to do.

seanwiseman: I can't seem to get this 'not' to return `[]` with some bits I have thrown at it. What is the data fixture you are testing with?

PR #63, opened 2017-12-14
Handle affiliations and email addresses being deleted

thewilkybarkid:


profiles/events.py
```diff
@@ -57,7 +57,7 @@ def event_handler(sender: Any, changes: List[Tuple[db.Model, str]]) -> None:
             if isinstance(instance, Profile):
                 ids.append(instance.id)
             if isinstance(instance, (Affiliation, EmailAddress)):
-                ids.append(instance.profile.id)
+                ids.append(instance.profile_id)
```
seanwiseman: (From previous discussion)

You lose the Profile model relationship reference from the EmailAddress or Affiliation object instance(s) passed to the event handlers on deletion, whereas they are available on update and insert. That's why 'instance.profile` was `None` and `instance.profile_id` was available.

test/events/test_send_update_events.py
```diff
@@ -52,6 +52,28 @@ def test_it_will_send_event_for_affiliation_insert(mock_publisher: MagicMock, pr
...
 
 
+def test_it_will_send_event_for_affiliation_deletion(mock_publisher: MagicMock, profile: Profile,
+                                                     session: scoped_session):
+    affiliation = Affiliation('1', Address(countries.get('gb'), 'City'), 'Organisation', Date(2017))
+
+    profile.add_affiliation(affiliation)
+    session.add(profile)
+
+    with patch('profiles.orcid.request'):
```
giorgiosironi: two lines that need to be extracted given all the tests use them, and the `patch` distracts from the intent of the test

seanwiseman: You are right, this is what I ended up doing in the api tests:
https://github.com/elifesciences/profiles/blob/3e9d901b1fd208acc02e9d782fc3be2405aae114/test/test_api.py#L11

Needs to be available as a fixture now that it has use in other files.

profiles/events.py
giorgiosironi: is `profile_id` also magically set by the ORM? I can't see an assignment for it

thewilkybarkid: Yep.

thewilkybarkid:
@giorgiosironi Good to go?

PR #64, opened 2017-12-14
Catch any exception raised by an event handler

thewilkybarkid:
Bad things seem to happen if an event handler raises any exception, it gets disconnected and the database transaction doesn't complete, requiring a restart.

### DIFF
--- a/profiles/utilities.py
+++ b/profiles/utilities.py
@@ -4,7 +4,8 @@ import random
 import string
 from datetime import datetime
 from functools import wraps
-from typing import Callable
+from logging import Logger
+from typing import Any, Callable
 
 from elife_api_validator import SCHEMA_DIRECTORY
 from flask import request
@@ -13,6 +14,20 @@ import pendulum
 from werkzeug.wrappers import Response
 
 from profiles.exceptions import SchemaNotFound
+
+
+def catch_exceptions(logger: Logger) -> Callable[..., Any]:
+    def outer_wrapper(func: Callable[..., Any]) -> Callable[..., Any]:
+        @wraps(func)
+        def wrapper(*args, **kwargs) -> Any:
+            try:
+                return func(*args, **kwargs)
+            except BaseException as exception:
+                logger.exception(exception)
+
+        return wrapper
+
+    return outer_wrapper
 
 
 def cache(allow_revalidation: bool = True) -> Callable[..., Response]:
--- a/profiles/utilities.py
+++ b/profiles/utilities.py
@@ -22,7 +22,7 @@ def catch_exceptions(logger: Logger) -> Callable[..., Any]:
         def wrapper(*args, **kwargs) -> Any:
             try:
                 return func(*args, **kwargs)
-            except BaseException as exception:
+            except Exception as exception:  # pylint: disable=broad-except
                 logger.exception(exception)
 
         return wrapper
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -7,7 +7,7 @@ from unittest.mock import MagicMock
 from _pytest.fixtures import FixtureRequest
 from flask import Flask
 from flask.testing import FlaskClient
-from flask_sqlalchemy import SQLAlchemy
+from flask_sqlalchemy import SQLAlchemy, models_committed
 from hypothesis import settings as hyp_settings
 from hypothesis.configuration import set_hypothesis_home_dir
 from itsdangerous import URLSafeSerializer
@@ -162,6 +162,17 @@ def orcid_token() -> OrcidToken:
 @fixture
 def orcid_tokens() -> SQLAlchemyOrcidTokens:
     return SQLAlchemyOrcidTokens(db)
+
+
+@fixture
+def registered_handler_names():
+    handler_names = []
+    for id_, recv in models_committed.receivers.items():  # pylint:disable=unused-variable
+        try:
+            handler_names.append(recv.__name__)
+        except AttributeError:
+            pass
+    return handler_names
 
 
 @fixture()
--- a/test/events/test_maintain_orcid_webhook.py
+++ b/test/events/test_maintain_orcid_webhook.py
@@ -1,4 +1,4 @@
-from typing import Dict
+from typing import Dict, List
 from unittest.mock import MagicMock, patch
 
 from flask_sqlalchemy import models_committed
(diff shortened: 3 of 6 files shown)


thewilkybarkid:
Not sure there's a way to test the event handler catch any exception, since 'normal' problems are already caught.

profiles/utilities.py
```diff
@@ -15,6 +16,20 @@
 from profiles.exceptions import SchemaNotFound
 
 
+def catch_exceptions(logger: Logger) -> Callable[..., Any]:
+    def outer_wrapper(func: Callable[..., Any]) -> Callable[..., Any]:
+        @wraps(func)
+        def wrapper(*args, **kwargs) -> Any:
+            try:
+                return func(*args, **kwargs)
+            except BaseException as exception:
```
seanwiseman: Catching `BaseException` will also catch system exiting exceptions. Blindly catching these can be dangerous, could we not use `Exception` so at least its just catching non-system exit exceptions?

profiles/utilities.py
```diff
@@ -15,6 +16,20 @@
 from profiles.exceptions import SchemaNotFound
 
 
+def catch_exceptions(logger: Logger) -> Callable[..., Any]:
```
seanwiseman: 👍 for decorator implementation.

profiles/utilities.py
thewilkybarkid: Would any non-`Exception` crash the application and cause it to restart?

thewilkybarkid: Hang on, misread. So any exception that doesn't extend `Exception` is definitely going to be a shutdown? In that case, we can ignore those.

thewilkybarkid:
Can't see a way to test the event handlers.

giorgiosironi:
Does this affect subsequent requests because there is only one worker process?
```
26749 ?        Ss     0:01 /srv/profiles/venv/bin/uwsgi --enable-threads --ini /srv/profiles/uwsgi.ini
26759 ?        Sl     0:01  \_ /srv/profiles/venv/bin/uwsgi --enable-threads --ini /srv/profiles/uwsgi.ini
```
So `26759` gets a broken connection and has to serve every following request?

If it gets that bad, is there a way to exit/crash? uwsgi will respawn a fresh process with a good connection. We catch exceptions in one place now but there could always be another place where they escape.

seanwiseman:
I will see if I can get tests working with exceptions being raised from within the handlers.

seanwiseman:
I have added some tests to `send_update_events` that show that an `Exception` will bubble up and cause badness when I mock out `@catch_exception` to do nothing, and then another to show that with `@catch_exception` in place, raising an exception from within the handler does not cause a kaboom.

I tried to attach to the actual logger instance that's set on the handler decorator to get the output but was not successful. Also quickly tried [pytest-catchlog](https://github.com/eisensheng/pytest-catchlog) but was not successful.

I did not get the same results performing the same methods on `maintain_orcid_webhook`, so will carry on with that now.

test/events/test_maintain_orcid_webhook.py
```diff
@@ -90,5 +96,50 @@ def test_it_ignores_other_models_being_committed(orcid_token: OrcidToken,
...
-    assert 'webhook_maintainer' in registered_handler_names
+@patch('profiles.events.catch_exceptions')
+def test_exception_not_handled_if_catch_decorator_is_removed(mock_catch: MagicMock,
+                                                             profile: Profile,
+                                                             orcid_config: Dict[str, str],
+                                                             mock_orcid_client: MagicMock,
+                                                             session: scoped_session,
+                                                             url_safe_serializer:
+                                                             URLSafeSerializer):
+    with pytest.raises(Exception):
```
thewilkybarkid: This shouldn't cover the whole thing? `session.add()` seems to throw an exception so the remaining isn't tested.

thewilkybarkid: This shouldn't cover the whole test? Removing it sees an earlier exception raised.

test/events/test_maintain_orcid_webhook.py
```diff
@@ -90,6 +96,49 @@ def test_it_ignores_other_models_being_committed(orcid_token: OrcidToken,
...
-    registered_handler_names = [recv.__name__ for id_, recv in models_committed.receivers.items()]
-    assert 'webhook_maintainer' in registered_handler_names
+@patch('profiles.events.catch_exceptions')
+def test_exception_not_handled_if_catch_decorator_is_removed(profile: Profile,
+                                                             orcid_config: Dict[str, str],
+                                                             mock_orcid_client: MagicMock,
+                                                             session: scoped_session,
+                                                             url_safe_serializer: URLSafeSerializer
+                                                             ):
+    with pytest.raises(Exception):
```
thewilkybarkid: This shouldn't cover the whole test? Removing it sees an earlier exception raised.

seanwiseman: With `mock_catch` removed `session.add(profile)` will throw an error because the `@patch` decorator needs to assign itself to something, and it will be the first arg (`profile` with it removed). So it was trying to add a MagicMock to the session. I will move the exception detection further down, the role of the test was to show it handles unknown exceptions so it did work but granted it should either detect the last part of the test.

thewilkybarkid: Oh I see; I get the same error at a different at a later point with it present (after removing `with pytest.raises(Exception)`).

test/conftest.py
```diff
@@ -164,6 +164,17 @@ def orcid_tokens() -> SQLAlchemyOrcidTokens:
     return SQLAlchemyOrcidTokens(db)
 
 
+@fixture
+def registered_handler_names():
```
thewilkybarkid: Can you extract this change into a new PR? It's unrelated to this change and keeps happening on other builds. We should get it in first.

test/events/test_maintain_orcid_webhook.py
seanwiseman: I am removing the two tests that basically remove the catch_exceptions decorator and test for failures. I have wrestled with mock based errors and they are not working as needed in this scenario. We have the tests for the actual decorator and we also have a test now for each handler raising and handling exceptions, I think it covers it.

test/conftest.py
thewilkybarkid: Unless we're about to merge!

thewilkybarkid:
@giorgiosironi Any final thoughts?

test/conftest.py
thewilkybarkid: (Do separate PRs in the future though, keeps unrelated changes separate and is clearer.)

thewilkybarkid:
👍

PR #65, opened 2017-12-14
Handle missing affiliation start dates

thewilkybarkid:
While looking through the logs I saw that there's errors when missing an affiliation start date (@NickDuf's profile). Looking at the XSD for the API, it's possible for it to be missing. However, through the ORCID UI it's a required field, so this might be for legacy data.

This should handle it correctly.


thewilkybarkid:
Just spotted that the `start_year` column is `not null`.

thewilkybarkid:
Good to go?

PR #66, opened 2017-12-19
Add proofreader

seanwiseman:
This replaces the calls to `flake8` and `pylint` with `proofreader`. As this project currently has its own overrides the default configurations will be overridden. 

This means functionally nothing has changed but it brings the project closer in line with how other current and future Python-based projects are to be structured, at least in terms of linting and pre-test checks.

### DIFF
--- a/project_tests.sh
+++ b/project_tests.sh
@@ -3,6 +3,7 @@ set -e
 
 source venv/bin/activate
 
-pylint --reports=n manage.py profiles/ test/*.py
-flake8 manage.py profiles/ test/
+pip install proofreader==0.0.2
+
+python -m proofreader manage.py profiles/ test/
 python -m pytest --junitxml=build/pytest.xml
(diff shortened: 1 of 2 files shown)


project_tests.sh
```diff
@@ -3,6 +3,7 @@ set -e
 
 source venv/bin/activate
 
-pylint --reports=n manage.py profiles/ test/*.py
-flake8 manage.py profiles/ test/
+pip install proofreader==0.0.2
```
thewilkybarkid: This means that doing a `pip freeze` will see it added to `requirements.txt`?

giorgiosironi: If you use `project_tests.sh` locally I think so, usually it is called by Jenkins only.

seanwiseman: Yes, in local development you would then see it added if you use `project_tests.sh`. 

Though `pip freeze` is clunky as it expands sub-dependencies of actual dependencies, really we should be managing the requirement files a bit more tightly (I am working on this) for example currently there are x60 dependencies listed in the `requirements.txt`.

thewilkybarkid: Just checking (something to watch out for).

Definitely not a fan of the current `requirements.txt`.

PR #67, opened 2017-12-19
Add test coverage

seanwiseman:
Added test coverage report generation and integration with [coveralls](https://coveralls.io/github/elifesciences/profiles).

I have disabled auto PR comments (so no spamming) and opted to not have this be an active check on PRs like we do in our libraries. This can be changed if need be in the repo settings and the failure percentage for a decrease can be configured in coveralls.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,2 @@
+[run]
+source = profiles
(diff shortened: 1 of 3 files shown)


seanwiseman:
Good to go @thewilkybarkid ?

PR #68, opened 2018-01-10
Add create profile cli command

seanwiseman:
Primary use is for the ingestion of existing `{name: email}` pairs from other services.

### DIFF
--- a/profiles/cli.py
+++ b/profiles/cli.py
@@ -1,11 +1,13 @@
 import logging
 from abc import ABC
-from typing import Dict
+from typing import Dict, List
 
 from flask import url_for
-from flask_script import Command as BaseCommand
+from flask_script import Command as BaseCommand, Option
 from itsdangerous import URLSafeSerializer
 
+from profiles.exceptions import ProfileNotFound
+from profiles.models import Name, Profile
 from profiles.orcid import OrcidClient
 from profiles.repositories import Profiles
 from profiles.types import CanBeCleared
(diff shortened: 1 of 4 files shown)


profiles/cli.py
```diff
@@ -36,6 +38,44 @@ def run(self) -> None:
...
+    # pylint: disable=method-hidden,arguments-differ
+    def run(self, name, email) -> None:
+        # pylint is complaining about arguments differing from base class
+        # though this is the recommended way to define the option args in the
+        # run() definition...
+        if name and email:
+            try:
+                profile = self.profiles.get_by_email_address(email)
+            except ProfileNotFound:
+                profile = Profile(self.profiles.next_id(), Name(name.strip('"')))
```
giorgiosironi: strange that the `strip()` is necessary?

profiles/cli.py
```diff
@@ -36,6 +38,44 @@ def run(self) -> None:
...
+                self.profiles.add(profile)
+                profile.add_email_address(email=email, restricted=True)
+
+            return profile.id, email
```
giorgiosironi: when is the transaction committed? Would be ideal to do so before returning the id to the user (otherwise a transaction failure would make that `id` vanish)

profiles/cli.py
seanwiseman: It was due to the `name` argument having spaces in it which would then require the value to be wrapped in quotes on the command line:

`python manage.py create-profile --name "Sean Wiseman" -- email test@test.com`

It passes `name` as `"Sean Wiseman"` hence the strip. I have since found how to let the `Option` class do this for you. Change applied.

profiles/cli.py
seanwiseman: We are using `SQLALCHEMY_COMMIT_ON_TEARDOWN = True` by default in the config. We, therefore, do not manually commit the session anywhere in the app. As the command supports re running with the same data (perhaps due to an error?) is this OK for now?

PR #69, opened 2018-01-30
Update name value via cli command if existing profile found

seanwiseman:
This can be reverted after if we do not want to persist this functionality.


thewilkybarkid:
Could keep it if there's no ORCID?

seanwiseman:
I think its probably safer to revert it after its use, so it is used only to create new profiles. We could add some more logic to check ORCID exists etc but I don't see a use case at the moment.

seanwiseman:
Yes, tested manually.

PR #70, opened 2018-02-02
Stop non-5xx responses being logged as errors

thewilkybarkid:
Don't know if this'll stop 400 responses from appearing in New Relic, but we seem to be logging them as `ERROR` level.

### DIFF
--- a/profiles/api/errors.py
+++ b/profiles/api/errors.py
@@ -33,7 +33,10 @@ def client_error_handler(exception: ClientError) -> Response:
 
 
 def http_error_handler(exception: HTTPException) -> Response:
-    LOGGER.exception(exception)
+    if exception.code < 500:
+        LOGGER.info(exception)
+    else:
+        LOGGER.exception(exception)
 
     return _handle_http_error(exception)
 


giorgiosironi:
Should not influence New Relic, safekeeping here of what I discovered yesterday:
```
https://docs.newrelic.com/docs/agents/python-agent/getting-started/instrumented-python-packages states

Exception logging is provided for:

    Uncaught exceptions which would otherwise generate 500 responses from Flask.

but they are in fact 400s:

10.0.2.195 - - [01/Feb/2018:14:32:23 +0000] "POST /oauth2/token HTTP/1.0" 400 76 "-" "GuzzleHttp/6.2.1 curl/7.35.0 PHP/7.0.23-1+ubuntu14.04.1+deb.
sury.org+1" 1.475

These should be coming from the New Relic instrumentation that is inserted in Flask (exceptions do not go up to the uwsgi daemon).

https://docs.newrelic.com/docs/agents/python-agent/configuration/python-agent-configuration#error-ignore and the following ignore_status_codes should do?
```

PR #73, opened 2018-02-07
ELPP-3316 Build a container image for development usage

giorgiosironi:
```
docker-compose build
docker-compose up
```
Will then be able to answer `/ping`:
```
$ curl -v localhost:8080/ping
> GET /ping HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.47.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Server: nginx/1.13.7
< Date: Thu, 08 Feb 2018 14:16:06 GMT
< Content-Type: text/plain; charset=UTF-8
< Content-Length: 4
< Connection: keep-alive
< Cache-Control: must-revalidate, no-cache, no-store, private
< 
pong
```

We don't have a database yet, so the container image can't run tests for now.


install.sh
```diff
@@ -16,6 +16,8 @@ if pip list | grep api-validator-python; then
 fi
 
 pip install --requirement requirements.txt
-python manage.py db upgrade
+if [ -z "$PROFILES_SKIP_DB" ]; then
+    python manage.py db upgrade
```
thewilkybarkid: Probably better to separate this as a step generally. Installing the dependencies shouldn’t require a functioning (ie configured) app.

giorgiosironi: Yes, eventually these will be two different scripts. I think one run at container creation time, and the other to be run by the `cli` container as a custom entrypoint where needed (during database setup).

giorgiosironi:
The error on installation is
```
Installing collected packages: alembic, bus-sdk-python, api-validator-python, astroid, attrs, blinker, boto3, botocore, certifi, chardet, click, coverage, docutils, Faker, flake8, flake8-import-order, Flask, Flask-Migrate, Flask-Script, Flask-SQLAlchemy, freezegun, hypothesis, idna, iso3166, isort, itsdangerous, Jinja2, jmespath, jsonschema, lazy-object-proxy, Mako, MarkupSafe, mccabe, newrelic, pendulum, psycopg2, py, pycodestyle, pyflakes, pylint, pytest, pytest-socket, python-dateutil, python-editor, python-json-logger, pytz, pytzdata, PyYAML, requests, requests-mock, retrying, s3transfer, six, SQLAlchemy, text-unidecode, tzlocal, urllib3, uWSGI, Werkzeug, wrapt

  Running setup.py install for alembic

    

    warning: no files found matching '*.jpg' under directory 'docs'

    warning: no files found matching '*.sty' under directory 'docs'

    warning: no files found matching '*.dat' under directory 'tests'

    no previously-included directories found matching 'docs/build/output'

    Traceback (most recent call last):

      File "<string>", line 1, in <module>

      File "/tmp/pip-build-ahf4t7x6/alembic/setup.py", line 73, in <module>

        'console_scripts': ['alembic = alembic.config:main'],

      File "/usr/local/lib/python3.5/distutils/core.py", line 148, in setup

        dist.run_commands()

      File "/usr/local/lib/python3.5/distutils/dist.py", line 955, in run_commands

        self.run_command(cmd)

      File "/usr/local/lib/python3.5/distutils/dist.py", line 974, in run_command

        cmd_obj.run()

      File "/srv/profiles/venv/lib/python3.5/site-packages/setuptools/command/install.py", line 61, in run

        return orig.install.run(self)

      File "/usr/local/lib/python3.5/distutils/command/install.py", line 551, in run

        self.run_command(cmd_name)

      File "/usr/local/lib/python3.5/distutils/cmd.py", line 313, in run_command

        self.distribution.run_command(command)

      File "/usr/local/lib/python3.5/distutils/dist.py", line 973, in run_command

        cmd_obj.ensure_finalized()

      File "/usr/local/lib/python3.5/distutils/cmd.py", line 107, in ensure_finalized

        self.finalize_options()

      File "/srv/profiles/venv/lib/python3.5/site-packages/setuptools/command/install_lib.py", line 16, in finalize_options

        self.set_undefined_options('install',('install_layout','install_layout'))

      File "/usr/local/lib/python3.5/distutils/cmd.py", line 290, in set_undefined_options

        setattr(self, dst_option, getattr(src_cmd_obj, src_option))

      File "/usr/local/lib/python3.5/distutils/cmd.py", line 103, in __getattr__

        raise AttributeError(attr)

    AttributeError: install_layout

    Complete output from command /srv/profiles/venv/bin/python3.5 -c "import setuptools, tokenize;__file__='/tmp/pip-build-ahf4t7x6/alembic/setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" install --record /tmp/pip-q0azy8el-record/install-record.txt --single-version-externally-managed --compile --install-headers /srv/profiles/venv/include/site/python3.5:

    running install
```
(from https://alfred.elifesciences.org/blue/organizations/jenkins/pull-requests-projects%2Fprofiles/detail/PR-73/6/pipeline)

Upgrading `setuptools` globally for Python 3.5 doesn't change it. Upgrading/installing it in the virtualenv before `pip install` works around it.

seanwiseman:
From the looks of it that is all you can do (upgrade `setuptools`), it's strange the global upgrade does not translate to the new virtualenv inheriting the upgraded version of `setuptools`.

`python3 -m easy_install --version`
will output the current setuptools version:
`setuptools 36.5.0 from /usr/local/lib/python3.5/site-packages (Python 3.5)`

Would be good to see what the version is both and after a global upgrade and what version the virtualenv has before upgrading it.

giorgiosironi:
Global version:
```
setuptools 38.5.1 from /usr/local/lib/python3.5/site-packages (Python 3.5)
```
Upgraded global version: 
```
root@a4c0a165792c:/srv/profiles# pip install -U setuptools
Requirement already up-to-date: setuptools in /usr/local/lib/python3.5/site-packages
root@a4c0a165792c:/srv/profiles# python3.5 -m easy_install --version
setuptools 38.5.1 from /usr/local/lib/python3.5/site-packages (Python 3.5)
```
Virtualenv initial version (??):
```
elife@1574e22772f0:/srv/profiles$ venv/bin/python3.5 -m easy_install --version
setuptools 5.5.1
```

Virtualenv upgraded version:
```
$ venv/bin/python3.5 -m easy_install --version
setuptools 38.5.1 from /srv/profiles/venv/lib/python3.5/site-packages (Python 3.5)
```

TL;DR: creating a virtualenv does not pick up the global `setuptools`:
```
elife@91ab547c7e95:/srv/profiles$ python3 -m easy_install --version
setuptools 38.5.1 from /usr/local/lib/python3.5/site-packages (Python 3.5)
elife@91ab547c7e95:/srv/profiles$ virtualenv --python=python3.5 venv
Running virtualenv with interpreter /usr/local/bin/python3.5
Using base prefix '/usr/local'
New python executable in venv/bin/python3.5
Also creating executable in venv/bin/python
Installing setuptools, pip...done.
elife@91ab547c7e95:/srv/profiles$ venv/bin/python3 -m easy_install --version
setuptools 5.5.1
```

giorgiosironi:
This virtualenv is very old, I'll try to install a newer one:
```
# dpkg -l virtualenv | grep virtualenv
ii  virtualenv     1.11.6+ds-1  all          Python virtual environment creator
```

seanwiseman:
I assume then `setuptools 5.5.1` is coming from the original python2.7 installation somehow??

Running `python -m easy_install --version` should logically return `setuptools 5.5.1 ...`

Also what is the virtualenv version? as I assume its install via python 2.7?

giorgiosironi:
Changed from the `virtualenv` from the Debian repositories to one installed via `pip`, which is much better:
```
root@592ce42fa2b3:/srv/profiles# pip install virtualenv
Collecting virtualenv
  Downloading virtualenv-15.1.0-py2.py3-none-any.whl (1.8MB)
    100% |████████████████████████████████| 1.8MB 863kB/s 
Installing collected packages: virtualenv
Successfully installed virtualenv-15.1.0
root@592ce42fa2b3:/srv/profiles# /usr/local/bin/virtualenv --version
15.1.0
root@592ce42fa2b3:/srv/profiles# /usr/local/bin/virtualenv --python=python3.5 venv
Running virtualenv with interpreter /usr/local/bin/python3.5
Using base prefix '/usr/local'
New python executable in /srv/profiles/venv/bin/python3.5
Also creating executable in /srv/profiles/venv/bin/python
Installing setuptools, pip, wheel...done.
root@592ce42fa2b3:/srv/profiles# venv/bin/python -m easy_install --version
setuptools 38.5.1 from /srv/profiles/venv/lib/python3.5/site-packages (Python 3.5)
```

giorgiosironi:
Now the installation works fine, thanks for your help @seanwiseman

Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM python:3.5.5-jessie
```
giorgiosironi: Like for `annotations`, we start from a Debian image; specialized for Python 3.5 in this case

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+RUN pip install virtualenv
+RUN pip install -U setuptools
+# END base image
+
+USER elife
+ENV PROJECT_FOLDER=/srv/profiles
+RUN mkdir ${PROJECT_FOLDER}
+WORKDIR /srv/profiles
+COPY --chown=elife:elife install.sh requirements.txt /srv/profiles/
+RUN PROFILES_SKIP_DB=1 /bin/bash install.sh
```
giorgiosironi: bare minimum installation of dependencies, this layer should be cached despite changes to the code in `profiles/`

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+## yes this is how you copy directories
+COPY --chown=elife:elife manage.py /srv/profiles/
+COPY --chown=elife:elife migrations /srv/profiles
+COPY --chown=elife:elife profiles/ /srv/profiles/profiles
```
giorgiosironi: omitting tests/ and similar: that can be done in an extension `profiles_ci` image (that is never published)

Dockerfile.wsgi
```diff
@@ -0,0 +1,5 @@
+ARG commit=latest
+FROM elifesciences/profiles_cli:${commit}
+
+ENV LANG=en_US.UTF-8
+ENTRYPOINT ["venv/bin/uwsgi", "--ini=/srv/profiles/uwsgi.ini"]
```
giorgiosironi: By customizing the entrypoint and moving up the environment variable I suspect we could use just a single image rather than a separate one for the application server (`wsgi` here, `fpm` in the PHP case).

config/app.cfg
```diff
@@ -0,0 +1,24 @@
...
+db = postgresql://username:password@hostname/profiles
+server_name = profiles--dev
+scheme = http
+
+[logging]
+path = var/logs
+level = DEBUG
+
+[orcid]
+api_uri = https://api.sandbox.orcid.org
```
giorgiosironi: all these configurations are lifted by the `.dist` files, but they will change to point to other containers: `orcid-dummy` and [`postgres`](https://hub.docker.com/_/postgres/) (official image)

config/nginx-default.conf
```diff
@@ -0,0 +1,12 @@
+server {
+    server_name profiles--dev;
+    listen 80;
+
+    location / {
+        uwsgi_pass wsgi:9000;
```
giorgiosironi: The `nginx` and `profiles_wsgi` containers communicate over a port, so they do not depend on a shared filesystem.

config/nginx-default.conf
```diff
@@ -0,0 +1,12 @@
...
+        uwsgi_read_timeout 15s;
+        include /etc/nginx/uwsgi_params;
+        uwsgi_param HTTP_HOST profiles--dev;
```
giorgiosironi: Configuring this value correctly is key to avoid Flask rejecting requests.

docker-compose.yml
```diff
@@ -0,0 +1,30 @@
+version: '3'
```
giorgiosironi: Right now a single file, but it could diverge into `docker-compose.yml` (locally) and `docker-compose.ci.yml` (for the `ci` stage).

.dockerignore
```diff
@@ -0,0 +1,8 @@
+adr/
```
thewilkybarkid: `.adr-dir` too

docker-compose.yml
```diff
@@ -0,0 +1,30 @@
...
+            - wsgi
+    wsgi:
+        build: 
+            context: .
+            dockerfile: Dockerfile.wsgi
+            args:
+                commit: "${IMAGE_TAG}"
+        image: elifesciences/profiles_wsgi:${IMAGE_TAG}
+        volumes:
+            # TODO: this can't be done as-is, it would overwrite the venv/
```
thewilkybarkid: 😢 

We'll need more in `volumes` to allow editing the actual code?

Dockerfile
thewilkybarkid: This is ignoring the DB for now?

docker-compose.yml
giorgiosironi: When this is separated into two `dev` and `ci` files, I plan to add `profiles/`, `test/`, etc. as separate volumes

giorgiosironi: Another option is we could place the `venv` folder elsewhere, to be able to mount the whole folder. Same potentially applies to `var/` however.

Dockerfile
giorgiosironi: Yes, I am checking out how Init Containers work in Kubernetes (and whether they can be ported to docker-compose) to copy the pattern. We a standard way to make database migration happen on startup, rather than at image build time.

giorgiosironi: (because the `postgres` image itself would have a volume containing the data, so every new deployment needs to re-run the setup)

giorgiosironi:
@seanwiseman updated to build everything through `docker-compose`, in the end it was even simpler to stick to a single image rather than building separate `cli` and `wsgi` ones (see commit message).

giorgiosironi:
Follow steps in the PR description to build locally.

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+    mkdir -p /var/www && \
+    chown www-data:www-data /var/www
+
+RUN pip install -U virtualenv
+RUN pip install -U setuptools
+# END base image
+
+USER elife
+ENV PROJECT_FOLDER=/srv/profiles
+ENV LANG=en_US.UTF-8
```
thewilkybarkid: Should be in the base image?

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+RUN PROFILES_SKIP_DB=1 /bin/bash install.sh
+
+## yes this is how you copy directories
+COPY --chown=elife:elife manage.py /srv/profiles/
+COPY --chown=elife:elife migrations /srv/profiles/
+COPY --chown=elife:elife profiles/ /srv/profiles/profiles
+
+RUN mkdir -p var/logs
+USER root
+RUN chown www-data:www-data var/logs
```
thewilkybarkid: Could be combined with the `mkdir`?

config/clients.yaml
```diff
@@ -0,0 +1,6 @@
+client-name:
+  client_id: client_id
+  client_secret: client_secret
+  redirect_uri: http://www.example.com/check
+
+# ...
```
thewilkybarkid: The comment was in the original file to demonstrate repetition, not needed here.

Dockerfile
giorgiosironi: the `mkdir` is executed as `elife` so that it leaves `var/` owned as `elife:elife`, while `var/logs/` becomes owned by `www-data:www-data`; change of ownership requires root

thewilkybarkid: `var` is in the repo but doesn't have a `COPY`, for clarity either just `mkdir -p var` as `elife` or add one (`COPY`) above and move the `mkdir` here. That would make it a bit clearer.

giorgiosironi: `COPY` is dangerous since it may pick up local log files and bake them into the image, whereas the folder should be empty by default; so I'll move the `var/` creation up

seanwiseman:
Will the database be added to the `docker-compose` file in this PR?

giorgiosironi:
> Will the database be added to the docker-compose file in this PR?

No as it was expanding too much, will be the next slice (running tests rather than only `/ping`)

giorgiosironi:
So, going to merge this and start with the next slice.

PR #74, opened 2018-02-13
Add postgres container, run ci

giorgiosironi:
Currently have to run

```
docker-compose exec wsgi venv/bin/python manage.py db upgrade
```

manually after `docker-compose up`, to create the schema.


giorgiosironi:
Can wait for the database initialization now, and run most tests.

However when running (after `up`):
```
docker-compose -f docker-compose.ci.yml run ci venv/bin/coverage run -m pytest
```
I get some errors
```
E           PermissionError: [Errno 13] Permission denied: '/srv/profiles/build/hypothesis/db/cd52e1f438478aab/6664ea9edbe52561.6e0dfa0ada81c086a1
381cffde349068'
```
and I'm trying to make Hypothesis (not Hypothes.is) write them elsewhere.

giorgiosironi:
Files are created by `www-data`, but they are `755` rather than `775` so the `elife` user running the Jenkins processes can't delete them at the moment, failing the build.

giorgiosironi:
So, waiting for the `postgres` container to come up is not enough as the schema (e.g. `profiles` table) won't be there yet.
What we really need is to wait for the `migrate` container to exit with 0 status; then we can smoke test (or in the future even run project tests that involve Postgres).

giorgiosironi:
It seems the `coveralls` python library hardcodes execution of a git command
```
17:12:03 -------------- generated xml file: /srv/profiles/build/pytest.xml --------------
17:12:03 ========================= 199 passed in 27.01 seconds ==========================
17:12:04 Submitting coverage to coveralls.io...
17:12:04 Traceback (most recent call last):
17:12:04   File "/srv/profiles/venv/bin/coveralls", line 11, in <module>
17:12:04     sys.exit(main())
17:12:04   File "/srv/profiles/venv/lib/python3.5/site-packages/coveralls/cli.py", line 62, in main
17:12:04     result = coverallz.wear()
17:12:04   File "/srv/profiles/venv/lib/python3.5/site-packages/coveralls/api.py", line 108, in wear
17:12:04     json_string = self.create_report()
17:12:04   File "/srv/profiles/venv/lib/python3.5/site-packages/coveralls/api.py", line 126, in create_report
17:12:04     data = self.create_data()
17:12:04   File "/srv/profiles/venv/lib/python3.5/site-packages/coveralls/api.py", line 177, in create_data
17:12:04     self._data.update(self.git_info())
17:12:04   File "/srv/profiles/venv/lib/python3.5/site-packages/coveralls/api.py", line 217, in git_info
17:12:04     rev = run_command('git', 'rev-parse', '--abbrev-ref', 'HEAD').strip()
17:12:04   File "/srv/profiles/venv/lib/python3.5/site-packages/coveralls/api.py", line 264, in run_command
17:12:04     'STDERR: "%s"' % (cmd.returncode, stdout, stderr))
17:12:04 AssertionError: command return code 128, STDOUT: "b''"
17:12:04 STDERR: "b'fatal: Not a git repository (or any of the parent directories): .git\n'"
```
which fails inside the Docker container, as there is no `.git` folder.

giorgiosironi:
It is a known problem:
https://github.com/coveralls-clients/coveralls-python/issues/174
I think I'll resort to mounting the `.git` folder as a volume.

.dockerignore
```diff
@@ -12,3 +12,5 @@ LICENSE
 README.md
 var/
 venv/
+**/__pycache__
+**/*.pyc
```
thewilkybarkid: Didn't `__pycache__` replace `.pyc` files?

wait_for_port
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -e
+
+port="${1:-80}"
+timeout="${2:-60}"
+host="${3:-localhost}"
+
+echo "Waiting for someone listening on port $port"
+timeout "$timeout" sh -c 'while ! nc -q0 -w1 -z '"$host"' '"$port"' </dev/null >/dev/null 2>&1; do sleep 1; done'
```
thewilkybarkid: Probably can be in the base image?

.dockerignore
seanwiseman: Well, in theory, all `.pyc` files should be in a `__pycache__` directory, but it was just a catch-all (you could generate the `.pyc` files manually and put them where you like).

.dockerignore
```diff
@@ -12,3 +12,5 @@ LICENSE
 README.md
 var/
 venv/
+**/__pycache__
```
thewilkybarkid: Best to keep trailing slashes for dirs?

.dockerignore
thewilkybarkid: Hmm, can't stop people doing silly things. 😉

thewilkybarkid: (After a bit of reading, Python 3.2+ only uses stuff inside `__pycache__` anyway?)

wait_for_port
giorgiosironi: It should definitely be (also took a look at other tools), but the base image doesn't exist yet.

.dockerignore
seanwiseman: If we are enforcing/advising to use 3.5+ then yer its fine to remove 👍

giorgiosironi: This base image is using 3.5.5, though this implies we will upgrade any 2.7 service to 3.5 before containerization.

seanwiseman: (You got me thinking, so I had to test this as it conflicted with what I thought was true)

Python 3.2+ can actually still use `.pyc` files from anywhere, it's just that Python uses `__pycache__` to manage them. Python won't generate them automatically anywhere other than a `__pycache__` (this is where they should be) but if you have valid `.pyc` elsewhere it can still be used.

This is not reason enough to warrant checking for them in this file (it's a rare edge case), it can rightfully be removed, but it is still a possibility. I could give you a whole directory of `.pyc` files that I have created and it would still run the project on your machine.

Thank you for questioning it, I have refreshed my memory of Python's cache checking order and management (a perfect start to a Monday 😁).

wait_for_port
giorgiosironi: (the `COPY` step is in the "move into base image" section)

giorgiosironi:
Going ahead with this, next step is tackle local usage.

PR #75, opened 2018-02-19
Local usage

giorgiosironi:
Running tests or running the application seems easy, but it remains to be seen how does this integrate with an IDE? That's the main point to try out.

You can obtain a virtualenv created inside a Docker container, but whether that will be portable to any host is unknown.


giorgiosironi:
Problem: `*.pyc` files can be picked up from the container when bind-mounting the source folders to run code with local modifications.
Possible sane/insane solution: https://github.com/inveniosoftware/invenio/commit/aaa9dd05dace04aec8eae6e09bb2365ff47a94a2

seanwiseman:
I can run the tests now using this method and it does not trigger the `.pyc` file issue.

I do however get presented with this message from `pytest` before running the tests:
```WARNING: assertions not in test modules or plugins will be ignored because assert statements are not executed by the underlying Python interpreter (are you using python -O?)``` (why yes we are, it knows 😱)

All tests are currently passing, but I do not know the possible future implications of the above message. Will dig a bit deeper.

giorgiosironi:
@lsh-0 started replacing assertions in code with https://github.com/elifesciences/builder/blob/master/src/buildercore/utils.py#L257-L261 and now I understand why

seanwiseman:
Ok now I understand.

thewilkybarkid:
Huh, same as PHP's `assert`.

giorgiosironi:
> Huh, same as PHP's assert.

Which results in everyone using `beberlei/assert` right?

seanwiseman:
I have tried using it for local dev and have been able to successfully bind to the remote interpreter from within the container launched by the `docker-compose.yml`.

One thing to note it the process of the tests running in this way (binding to the remote `virtualenv`) seems to take 2 to 3 times longer than just using a local `virtualenv` (30 seconds vs 60 - 90 seconds), seems minor but started to add up after a few changes. Not sure there is anything we can do about this as it's obviously just having to do more each time it inits.

The other thing is the clean up which we may be able to tackle. If I leave my test runner process in file watch mode (or if I run any command via `docker-compose`) I get a remnant of each run in the form of a container (profiles_ci_run_1, profiles_ci_run_2 ...). Just run it a couple of times then run `docker-compose down` and you will see all the instances it has to terminate. If I were to use it throughout the day without running `down` all the time I could end up with hundreds of these.

giorgiosironi:
> I have tried using it for local dev and have been able to successfully bind to the remote interpreter from within the container launched by the docker-compose.yml.

I too set up PyCharm to do this, but was unsure how to run the tests?

giorgiosironi:
> If I leave my test runner process in file watch mode (or if I run any command via docker-compose) I get a remnant of each run in the form of a container (profiles_ci_run_1, profiles_ci_run_2 ...)

Updated the command in the `README`, this avoids leaving containers in `docker container ls -a`'s list.

giorgiosironi:
> One thing to note it the process of the tests running in this way (binding to the remote virtualenv) seems to take 2 to 3 times longer than just using a local virtualenv (30 seconds vs 60 - 90 seconds)

I get ~17-18 seconds locally and ~20 seconds using the container, this could be related to the file sharing with the Linux VM on a Mac (I'll fiddle with a Mac to check). It doesn't seem to worsen with the source folders being mounted though.

seanwiseman:
> I get ~17-18 seconds locally and ~20 seconds using the container, this could be related to the file sharing with the Linux VM on a Mac (I'll fiddle with a Mac to check). It doesn't seem to worsen with the source folders being mounted though.

Just to clarify, I mean binding to the `ci` container's `virtualenv`. If I just run `docker-compose run --rm ci venv/bin/pytest` then its fine, no real change from running it locally. 

But If you truly want to use this to 'develop' locally (be able to debug etc) (with IDE integration) you need to bind to the `ci` container's `virtualenv`, which you can but you get the slightly longer feedback loop. 

I am not saying this is even something we can address at present, it's down to the tooling and the fact you are doing more than just running a local `virtualenv`. When weighing it up though, the benefit of being able to debug/run tests in a production like environment outweighs the slight additional time for the tests to run (in my opinion) 👍.

giorgiosironi:
> But If you truly want to use this to 'develop' locally (be able to debug etc) (with IDE integration) you need to bind to the ci container's virtualenv, which you can but you get the slightly longer feedback loop.

Not sure how that works (to then optimize it); since the host shouldn't be able to see the container's filesystem, it could imply it's running `python ...` commands repeatedly to extract information? I'll try to set it up too.

giorgiosironi:
Verified a slight overhead of PyCharm in running pytest on a Mac, 30 seconds rather than 25.

PR #76, opened 2018-02-19
New Relic support, but keeping it disabled by default

giorgiosironi:
Locally and in `ci`, New Relic can usually be turned off.

More environment variables are needed, but only if New Relic is actually used (e.g. license key).


giorgiosironi:
Coverage decreased, apparently.

giorgiosironi:
`profiles.factory` changes but it's still covered 100%, probably the fact that it became shorter brought down total coverage.

PR #78, opened 2018-02-20
Set up dummies: orcid-dummy, goaws

giorgiosironi:
Dummies are useful in testing environments: `dev`, `ci` (which is now very similar to `dev` apart from the separate `docker-compose.ci.yml` file), and `end2end`.


config/app.cfg
```diff
@@ -22,3 +22,6 @@ webhook_key = some_string
 region = us-east-1
 subscriber = 512686554592
 name = bus-profiles
+env = dev
+# TODO: not implemented yet?
+endpoint = http://goaws:4100
```
giorgiosironi: @seanwiseman you were testing locally with `goaws`, so how do we extend `bus-sdk-python` to accept the `endpoint[ url]` for local usage?

config/app.cfg
```diff
@@ -22,3 +22,6 @@ webhook_key = some_string
 region = us-east-1
 subscriber = 512686554592
 name = bus-profiles
+env = dev
```
giorgiosironi: I saw `env` in the current configuration provided by the formula, but maybe it's obsolete?

config/app.cfg
seanwiseman: All you need to do is pass in `endpoint_url` as a config override and the initialization is handled for you.
```
from elife_bus_sdk import get_publisher

config = {
    'region': 'us-east-2',
    'subscriber': '00000000000',       
    'name': 'profile',
    'endpoint_url': 'http://goaws:4100'  # or from a config variable etc
}

publisher = get_publisher(pub_type='sns', config=config)
```
This would give you an SNS 'publisher' pointed at your local `goaws` instance. This will be the same with SNS queues.

config/app.cfg
seanwiseman: It's used when generating the `arn`:

```
return 'arn:aws:sns:{region}:{subscriber}:{name}--{env}'.format(region=region,
                                                                subscriber=subscriber,
                                                                name=name,
                                                                env=env)
```

thewilkybarkid: That's `profiles.environment`.

giorgiosironi:
Ready for review now.

PR #83, opened 2018-02-27
Automatically read web and db hosts in smoke tests

giorgiosironi:
Started out to make them configurable, but it was more foolproof to just read the configuration to keep smoke tests easy to run from any user or automated process.

### DIFF
--- a/profiles/cli.py
+++ b/profiles/cli.py
@@ -20,8 +20,8 @@ class Command(ABC, BaseCommand):
         with app.app_context():
             return self.run(*args, **kwargs)
 
-"Allows calling a method on the Config object to print its result"
 class ReadConfiguration(Command):
+    "Allows calling a method on the Config object to print its result"
     NAME = 'read-configuration'
 
     def __init__(self, config) -> None:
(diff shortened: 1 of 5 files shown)


profiles/cli.py
```diff
@@ -20,6 +20,26 @@ def __call__(self, app=None, *args, **kwargs) -> None:
         with app.app_context():
             return self.run(*args, **kwargs)
 
+"Allows calling a method on the Config object to print its result"
```
seanwiseman: Should be under the class definition as the class doc string.

profiles/cli.py
```diff
@@ -20,6 +20,26 @@ def __call__(self, app=None, *args, **kwargs) -> None:
...
+        super(ReadConfiguration, self).__init__()
+        self.config = config
+
+    def get_options(self) -> List[Option]:
+        return [
+            Option('-s', '--method', dest='method', type=str),
+        ]
+
+    # pylint: disable=method-hidden
+    def run(self, method) -> None:
```
seanwiseman: You will need to add:
https://github.com/elifesciences/profiles/blob/fe32751d4769f76663207e0bb0fba5f34b2da34b/profiles/cli.py#L77

to avoid the same `pylint` issue I had with the last command added.

giorgiosironi:
Tests green, coveralls red; this code however will be run by every deploy.

PR #84, opened 2018-02-27
Use base image

giorgiosironi:
On `latest` as there is no published version yet. Depends on https://github.com/elifesciences/elife-base-images/pull/14


giorgiosironi:
I'll pin this before merging.

PR #85, opened 2018-02-28
Extract dockerCompose*() steps

giorgiosironi:
If you fancy seeing where the logic has gone, the custom steps are in https://github.com/elifesciences/elife-jenkins-workflow-libs/tree/master/vars


giorgiosironi:
Last checkpoint of the day probably: the JUnit test results can be retrieved from the container, but `JUnitResultArchiver` then had a `NullPointerException`.

PR #88, opened 2018-04-25
ELPP-3509 Improve images

thewilkybarkid:


Dockerfile.ci
```diff
@@ -1,19 +1,21 @@
...
-COPY --chown=elife:elife test/ /srv/profiles/test/
 COPY --chown=elife:elife \
     .coveragerc \
     .flake8 \
     project_tests.sh \
     .pylintrc \
     pytest.ini \
-    /srv/profiles/
-RUN mkdir build/ && chmod 777 build/
+    requirements-debug.txt
```
giorgiosironi: is this missing a final `\` but works anyway?

Dockerfile.venv
```diff
@@ -0,0 +1,14 @@
+FROM elifesciences/python:b22fbad8f66100d88cb4bd7c7d092b376a9e09bc
```
giorgiosironi: could also be named `.pip`, but in the end both `pip` and `virtualenv` are needed here

Dockerfile.venv
```diff
@@ -0,0 +1,14 @@
...
+WORKDIR ${PROJECT_FOLDER}
+
+USER root
+RUN chown --recursive elife:elife .
+
+USER elife
+
+RUN virtualenv venv
+
+COPY requirements.txt ./
```
giorgiosironi: There may be other steps not yet implemented, but if all this did was to create `venv/` the only difference with having it in the main `Dockerfile` would be the lack of a `requirements.txt` in the shipped one. Unless the base image of `Dockerfile` change to something smaller.

Dockerfile
```diff
@@ -1,20 +1,20 @@
...
 USER root
-RUN mkdir var/logs && chown www-data:www-data var/logs
+RUN mkdir -p var/logs && \
+    chown --recursive elife:elife . && \
+    chown www-data:www-data var/logs
+
+COPY --chown=elife:elife smoke_tests_wsgi.sh ./
+COPY --chown=elife:elife manage.py ./
+COPY --chown=elife:elife migrations/ migrations/
+COPY --from=venv --chown=elife:elife ${PROJECT_FOLDER}/venv/ venv/
```
giorgiosironi: we need to understand under which conditions a virtualenv is portable between different operating system images.

For example, there are binaries not necesaarily identical across architectures and that may be depending on shared objects libraries:
```
[giorgio@Newton:~/code/profiles/venv/bin]$ ls -1 . | grep -v py | xargs file | grep ELF
uwsgi:            ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 2.6.32, BuildID[sha1]=3ad79a028c941d6c3217ffbbcfffd5781b891926, not stripped
```

Plus there are a lot of symlinks to system-wide files:
```
[giorgio@Newton:~/code/profiles/venv]$ ls -l lib/python3.5/ | head
total 44
lrwxrwxrwx   1 giorgio giorgio    25 Feb 19 15:01 abc.py -> /usr/lib/python3.5/abc.py
lrwxrwxrwx   1 giorgio giorgio    28 Feb 19 15:01 base64.py -> /usr/lib/python3.5/base64.py
lrwxrwxrwx   1 giorgio giorgio    28 Feb 19 15:01 bisect.py -> /usr/lib/python3.5/bisect.py
lrwxrwxrwx   1 giorgio giorgio    33 Feb 19 15:01 _bootlocale.py -> /usr/lib/python3.5/_bootlocale.py
lrwxrwxrwx   1 giorgio giorgio    28 Feb 19 15:01 codecs.py -> /usr/lib/python3.5/codecs.py
lrwxrwxrwx   1 giorgio giorgio    30 Feb 19 15:01 collections -> /usr/lib/python3.5/collections
lrwxrwxrwx   1 giorgio giorgio    38 Feb 19 15:01 _collections_abc.py -> /usr/lib/python3.5/_collections_abc.py
lrwxrwxrwx   1 giorgio giorgio    47 Feb 19 15:01 config-3.5m-x86_64-linux-gnu -> /usr/lib/python3.5/config-3.5m-x86_64-linux-gnu
lrwxrwxrwx   1 giorgio giorgio    26 Feb 19 15:01 copy.py -> /usr/lib/python3.5/copy.py
```

This should all be fine if the base images are the same.

thewilkybarkid:
Tests work for me 🤔 /cc @seanwiseman

seanwiseman:
All tests pass for me also. Profiles Mac game is 💪

giorgiosironi:
`botocore.exceptions.NoRegionError: You must specify a region.` could point to the lack of a `AWS_DEFAULT_REGION` variable? Trying it out locally.

giorgiosironi:
`dev.env` specifies `AWS_DEFAULT_REGION` but tests are run directly through `docker run` without `docker-compose` in `dockerProjectTests`, will substitute it with `dockerComposeProjectTests`.

Jenkinsfile
```diff
@@ -10,20 +10,20 @@ elifePipeline {
...
             }
 
             stage 'Project tests', {
                 def coverallsToken = sh(script:'cat /etc/coveralls/tokens/profiles', returnStdout: true).trim()
                 withEnv(["COVERALLS_REPO_TOKEN=$coverallsToken"]) {
-                    dockerComposeProjectTests('profiles', commit, ['/srv/profiles/build/*.xml'])
+                    dockerComposeProjectTests 'profiles', commit
+                }
+                try {
+                    sh "IMAGE_TAG=${commit} docker-compose -f docker-compose.yml -f docker-compose.ci.yml up -d"
```
giorgiosironi: this is possibly missing a wait for the migrations container to have run its course

giorgiosironi:
I have updated the `dockerCompose*` steps to use the new standard of two `docker-compose*.yml` files, so that the various commands can be extracted across `annotations`|`profiles`|etc.

Dockerfile
```diff
@@ -1,28 +1,21 @@
-FROM elifesciences/python:b22fbad8f66100d88cb4bd7c7d092b376a9e09bc
+ARG image_tag=latest
+FROM elifesciences/profiles_venv:${image_tag} as venv
+ARG python_version
+FROM elifesciences/python:${python_version}
```
thewilkybarkid: Will be changed when https://github.com/elifesciences/elife-base-images/pull/20 is merged.

Dockerfile.venv
```diff
@@ -0,0 +1,21 @@
+ARG python_version
+FROM elifesciences/python:${python_version}
+
+ARG pipenv_dev_arg
+
+ENV LANG=C.UTF-8
```
thewilkybarkid: Not needed when https://github.com/elifesciences/elife-base-images/pull/20 is merged.

giorgiosironi:
```
FileNotFoundError: [Errno 2] No such file or directory: 'git': 'git'
```
The `ci` image needs to install `git` for coveralls to be able to detect metadata about the code?

thewilkybarkid:
Hmm, does being in `.dockerignore` stop a volume? How did it work before?

thewilkybarkid:
Hang on, it wants `git` not `.git`. The command?

Jenkinsfile
```diff
@@ -10,13 +10,13 @@ elifePipeline {
...
                 checkout scm
-                dockerComposeBuild(commit)
+                dockerComposeBuild commit
             }
 
             stage 'Project tests', {
                 def coverallsToken = sh(script:'cat /etc/coveralls/tokens/profiles', returnStdout: true).trim()
                 withEnv(["COVERALLS_REPO_TOKEN=$coverallsToken"]) {
-                    dockerComposeProjectTests('profiles', commit, ['/srv/profiles/build/*.xml'])
+                    dockerComposeProjectTests 'profiles', commit
```
giorgiosironi: seems to have lost artifacts as https://alfred.elifesciences.org/blue/organizations/jenkins/pull-requests-projects%2Fprofiles/detail/PR-88/14/tests is empty. Was this failing?

Dockerfile
```diff
@@ -1,28 +1,24 @@
...
-COPY --chown=elife:elife manage.py /srv/profiles/
-COPY --chown=elife:elife migrations/ /srv/profiles/migrations/
-COPY --chown=elife:elife profiles/ /srv/profiles/profiles/
-COPY --chown=elife:elife smoke_tests_wsgi.sh /srv/profiles/
-RUN mkdir /srv/profiles/var/
-
 USER root
-RUN mkdir var/logs && chown www-data:www-data var/logs
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    git \
```
giorgiosironi: I assumed this would only be needed in the `ci` image, although if we have to install `git` on every build it's easier to keep it here?

Dockerfile
giorgiosironi: Explicit strategy on this is to rely on the base images being identical? It can get out of sync given there is a base image in the middle

Jenkinsfile
thewilkybarkid: Looks like an oversight. Also looks like this file doesn't actually need to change? Reverting.

Dockerfile
thewilkybarkid: Wrong file. :doh:

thewilkybarkid:
Down to 88 MB from 358.

PR #89, opened 2018-04-25
ELPP-3509 Switch to Pipenv

thewilkybarkid:
Not yet working as VCS dependencies (ie `api-validator-python`) don't appear to work correctly. Unclear whether should be using virtualenv too.

An improvement on pip, but still problematic.

/cc @seanwiseman @giorgiosironi


Dockerfile
```diff
@@ -1,12 +1,20 @@
 FROM elifesciences/python:b22fbad8f66100d88cb4bd7c7d092b376a9e09bc
 
+ENV LANG=C.UTF-8
```
thewilkybarkid: click complained.

Pipfile
```diff
@@ -0,0 +1,46 @@
+[[source]]
+url = "https://pypi.python.org/simple"
+verify_ssl = true
+name = "pypi"
+
+[packages]
+alembic = "==0.9.5"
```
thewilkybarkid: These should be changed to `~=`.

profiles/utilities.py
```diff
@@ -75,25 +69,3 @@ def guess_index_name(name: str) -> str:
 
 def remove_none_values(items: dict) -> dict:
     return dict(filter(lambda item: item[1] is not None, items.items()))
-
-
-def validate_json(data: dict, schema_name: str, schema_dir: str = ''):
```
thewilkybarkid: Moved to keep `jsonschema` as a dev dependency.

Dockerfile
giorgiosironi: Not sure who click is, but I've seen this before e.g. in bot-lax-adaptor as environments like servers start from an empty set of environment variables that sometimes are instead set locally

Dockerfile
```diff
@@ -17,7 +25,7 @@ USER root
 RUN mkdir var/logs && chown www-data:www-data var/logs
 
 USER www-data
-CMD ["venv/bin/python"]
```
giorgiosironi: Docker containers should still use a virtualenv https://stackoverflow.com/a/33150800/91590 (whether it is through `pipenv` or `virtualenv` or another tool) to avoid interaction with the distribution packages; there's little else that run inside a container than the app itself, but `python` dependencies are surprisingly widespread in Linux images.

For example, you may find that installing a particular `requests` version in an app breaks [the system one](https://github.com/pypa/pipenv/blob/master/setup.py#L31), system tools then try to vendor it themselves (`pipenv` is doing this) shipping their own copy, in an arms race, etc.

Dockerfile
thewilkybarkid: http://click.pocoo.org/5/python3/#python-3-surrogate-handling

Dockerfile
```diff
@@ -17,7 +25,7 @@ USER root
...
+CMD [".venv/bin/python"]
```
thewilkybarkid: This is the remaining issue. You appear not to be able to control the name (https://github.com/pypa/pipenv/issues/1184), and it's different to what we use now. Going to see if a symlink works.

profiles/factory.py
```diff
@@ -42,7 +42,7 @@ def create_app(config: Config, clients: Clients) -> Flask:
 
     config_bus = dict(config.bus)
     config_bus['env'] = config.name
-    publisher = get_publisher(pub_name='profiles', config=config_bus)
+    publisher = get_publisher(config=config_bus)
```
thewilkybarkid: Update to the library caused this, do we need to set it elsewhere @seanwiseman?

seanwiseman: Ah yes. The instance storage and ref by name had to go, was not adding much anyway.

No, it does not need to be added anywhere else.

thewilkybarkid:
#88 will be updated to optimise some of this.

test/conftest.py
```diff
@@ -232,3 +235,26 @@ def wrapped() -> None:
             session.commit()
 
     return wrapped
+
+
+@fixture
+def validate_json():
```
seanwiseman: 👍

.pylintrc
```diff
@@ -1,5 +1,5 @@
 [MESSAGE CONTROL]
-disable = abstract-method,missing-docstring,redefined-outer-name,too-few-public-methods
+disable = abstract-method,inconsistent-return-statements,missing-docstring,redefined-outer-name,too-few-public-methods
```
thewilkybarkid: Pylint ended up upgraded, could downgrade or fix the problem.

test/conftest.py
```diff
@@ -232,3 +235,26 @@ def wrapped() -> None:
...
+        schema_path = os.path.join(schema_dir, '{}.json'.format(schema_name))
+
+        try:
+            with open(schema_path) as schema_file:
+                validate(data, schema=json.load(schema_file))
+            return True
+        except (FileNotFoundError, ValidationError):
+            # Need to re raise with schema/validation failure information,
+            # though as this will be replaced by api-validator-python
+            # leaving for now
```
thewilkybarkid: Can this be replaced now @seanwiseman?

seanwiseman: Yes, but the tests using it would need amending slightly as the `JSONResponseValidator` takes the HTTP response object as it takes into account the media type etc. I left validating `json` to the `jsonschema` lib as not to just write a wrapper around the func.

I would leave for another PR.

thewilkybarkid: I'll revert change to reduce the diff here since it can be replaced anyway.

.pylintrc
thewilkybarkid: Can't downgrade, version enforce by Proofreader.

seanwiseman: Whats causing the `inconsistent-return-statements` message?

thewilkybarkid: Can't remember which files, it's where not all functions have return statements.

install.sh
```diff
@@ -1,23 +0,0 @@
-#!/usr/bin/env bash
```
giorgiosironi: not used anymore by the formula https://github.com/elifesciences/profiles-formula/search?utf8=%E2%9C%93&q=install.sh&type=

project_tests.sh
```diff
@@ -2,8 +2,9 @@
 set -e
 
 source venv/bin/activate
+export PYTHONOPTIMIZE=
```
giorgiosironi: it was a dangerous option, we put it in to allow mixed host/container local usage as
- host would create `.pyc` files
- container would create `.pyo` files
and despite the host source mounted inside the container, they wouldn't interact or clash in file ownership. I don't think `project_tests.sh` is used much locally so it shouldn't be a problem.

PR #95, opened 2018-06-19
ELPP-3693 Run on dynamic Jenkins node

giorgiosironi:
Rather than on `containers--medium` which is manually managed

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -6,8 +6,7 @@ elifePipeline {
         commit = elifeGitRevision()
     }
 
-    elifeOnNode(
-        {
+    node('containers-jenkins-plugin') {
             stage 'Build images', {
                 checkout scm
                 dockerComposeBuild(commit)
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -6,8 +6,7 @@ elifePipeline {
         commit = elifeGitRevision()
     }
 
-    elifeOnNode(
-        {
+    node('containers-jenkins-plugin') {
             stage 'Build images', {
```
thewilkybarkid: Indentation could be reduced?

giorgiosironi:
Now green with the latest AMI (which ships coveralls credentials for everything) (but only `profiles` has some). I am looking for a way to streamline the AMI update, but that's a separate matter.

Jenkinsfile
giorgiosironi: Yes, I didn't want to pollute the diff too much

thewilkybarkid: That’s what `?w=1` is for. :)

PR #96, opened 2018-06-20
Reduce indentation of containers node stages

giorgiosironi:


giorgiosironi:
@seanwiseman it seems that coveralls is comparing the coverage result with the last run of any branch, rather than with `develop` which should be the reference. Possible?

seanwiseman:
The default branch is set to `develop` on the `coveralls` interface settings. Doesn't seem possible on paper, what is showing this?

giorgiosironi:
Having to merge this to build up on it another fix to the pushing of the image.

PR #100, opened 2018-11-02
updates requests to 2.20

lsh-0:


lsh-0:
gah! I'm really not getting pipenv. Could somebody else please update profiles to use `requests >= 2.20` before I replace it with a bogstandard `requirements.txt` file?

lsh-0:
It looks like @giorgiosironi is the defacto owner with the most number of recent commits. If there is no actual owner anymore and no objections I'm more than happy to replace pipenv.

de-code:
Should probably add `--ignore-pipfile` to the `pipenv` command in `Dockerfile.venv` to only use the dependencies as per the lock file.

giorgiosironi:
I'll retry this update to see what is going on with the other projects in the lock file.

> Should probably add --ignore-pipfile

Will investigate this too, was it related?

de-code:
> > Should probably add --ignore-pipfile
> 
> Will investigate this too, was it related?

I would say so, this build fails with `Your Pipfile.lock (c9fe50) is out of date. Expected: (de4a1b).`

With that option it won't check outdated dependencies but just adhere to the lock file.

giorgiosironi:
Uhuh: https://github.com/pypa/pipenv/issues/2665

giorgiosironi:
Which actually leads to https://github.com/pypa/pipenv/issues/2412

giorgiosironi:
You're right that there is a pre-existing problem with the Docker build itself because if I executed `pipenv graph` I see a newer version of `newrelic` wrt the current lock (`2.92.0.78`):
```
(venv) root@30a877876ea9:/srv/profiles# pipenv graph | grep newrelic
newrelic==2.106.1.88
```

giorgiosironi:
In a smaller example `pipenv sync` for installation and ~`pipenv update --keep-outdated requests` for updates of a single package seem to work.~ no they don't they update the `requests` dependencies but not the package itself.

giorgiosironi:
As far as I can understand the last word on this is https://github.com/pypa/pipenv/issues/966 which is open and locked due to heated discussion. I don't have much trust in pipenv but rather than jumping on other solutions like `poetry` I'll hold off and:

- fix the locking here (which seems to work)
- allow this to go ahead by updating the world rather than updating only `requests`

giorgiosironi:
I included the update, which is identical to this, in the other PR to simplify testing; let's go there.

lsh-0:
I had to go looking for it: https://github.com/sdispater/poetry

thanks guys, your work is appreciated

PR #101, opened 2018-11-02
Use Pipfile.lock and provide update instructions

giorgiosironi:
For https://github.com/elifesciences/profiles/pull/100 to go through

### DIFF
--- a/Dockerfile.venv
+++ b/Dockerfile.venv
@@ -1,6 +1,9 @@
 ARG python_version
 FROM elifesciences/python_3.6_pipenv:${python_version}
 
+# force latest pipenv, should be upgraded in base image
+RUN pip install --upgrade pipenv
+
 ARG pipenv_dev_arg
 
 ENV PROJECT_FOLDER=/srv/profiles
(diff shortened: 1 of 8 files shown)


Dockerfile.venv
```diff
@@ -10,4 +13,4 @@ RUN virtualenv venv
 ENV VIRTUAL_ENV=${PROJECT_FOLDER}/venv
 
 COPY Pipfile Pipfile.lock ./
-RUN pipenv install ${pipenv_dev_arg} --deploy
+RUN pipenv sync ${pipenv_dev_arg}
```
de-code: Actually it's documented here: https://pipenv.readthedocs.io/en/latest/advanced/

> pipenv install --ignore-pipfile is nearly equivalent to pipenv sync, but pipenv sync will never attempt to re-lock your dependencies as it is considered an atomic operation. pipenv install by default does attempt to re-lock unless using the --deploy flag.

(The docs for `--ignore-pipfile` should really point to that)

giorgiosironi: Yes, `sync` seemed marginally better so I went for that

giorgiosironi:
Some problems running the build because of Github downtime https://status.github.com/messages

giorgiosironi:
Now it looks some tests are vulnerable to DST as the diffs have a 1-hour lag? What next, earthquake? Flood? Locusts?

giorgiosironi:
Have to signoff now but I can continue cleaning up the project on Monday.

giorgiosironi:
You wish... Tuesday.

giorgiosironi:
Good fix, thank you

PR #103, opened 2018-12-11
Introduce development-oriented Docker container

GiancarloFusiello:
- added ability to run, test and debug code locally using new dev docker containers
- added more logging to improve debugging
- updated README with new make commands


Dockerfile.dev
```diff
@@ -0,0 +1,10 @@
+FROM python:3.6
```
giorgiosironi: consider https://github.com/elifesciences/elife-base-images/blob/master/Dockerfile.python_3.6#L1 which may be smaller in size, since it doesn't seem you need particular OS packages to be installed

Makefile
```diff
@@ -0,0 +1,23 @@
...
+	docker-compose -f docker-compose.dev.yml down
+
+test:
+	docker-compose -f docker-compose.dev.yml run --rm --service-ports web pytest -s
+
+debug:
+	docker-compose -f docker-compose.dev.yml run --rm --service-ports web
+
+clean:
+	-docker rm `docker ps -aq`
```
giorgiosironi: that's harsh, because it cleans the whole host from everything rather than what is related to the application; `docker-compose down -v` is usually enough?

giorgiosironi: (meaning the full 3 lines, the first one is not dangerous because it's only about containers rather than images and volumes)

Pipfile.lock
```diff
@@ -198,39 +225,39 @@
...
+                "sha256:b2c09359d6802279efb9efb3f91a9c94567151baee95175f9b637ea628f35244",
+                "sha256:ca7bc37b1efb7cc25271bf10f398462ed975d95259af1406d38fcb268466e34f",
+                "sha256:e64235d9013ebf6319cb9654e08f5066112c34d8c4cc41186254ab9c3d6d5b9b",
+                "sha256:ec9be679c0065667503851141c31fa699e1cc69ded3ba8e5d3673dd5a6eb1370",
+                "sha256:eca00d0f91fcb44d88b12f1fd16ad138e38fa07debb79587e2b7ff1fe80d72b9",
+                "sha256:f256e807b8b2b45b6af60d7f2bb5194aab2f4acc861241c4d8ef942a55f5030d",
+                "sha256:fce7612a3bd6a7ba95799f88285653bf130bd7ca066b52674d5f850108b2aec0"
+            ],
+            "index": "pypi",
+            "version": "==2.7.6.1"
```
giorgiosironi: as we can see, `pipenv` here upgraded random packages for no good reason; the only thing it needed to do was to install 2 new packages in the `dev` section

docker-compose.dev.yml
```diff
@@ -0,0 +1,37 @@
+version: '3'
```
giorgiosironi: probably at this point we should remove the existing `docker-compose.override.yml` in favor of this

docker-compose.dev.yml
```diff
@@ -0,0 +1,37 @@
...
+    env_file:
+      - dev.env
+
+  web:
+    build:
+      context: ./
+      dockerfile: Dockerfile.dev
+    working_dir: /srv/profiles
+    volumes:
+      - .:/srv/profiles
```
giorgiosironi: the custom `app.cfg` is not committed (it's in `.gitignore`) but it gets into the container from here, so unless it's very specific it's time to start committing with a name like `app.cfg.dev` or similar?

docker-compose.dev.yml
```diff
@@ -0,0 +1,37 @@
...
+    command: "python manage.py runserver --host=0.0.0.0"
+    ports:
+      - "5000:5000"
+    links:
```
giorgiosironi: [`links` is deprecated](https://docs.docker.com/compose/compose-file/#links) and `depends_on` should be used for the same intent here, which is to decide the order in which containers are started

profiles/config.py
```diff
@@ -1,3 +1,4 @@
+import os
```
giorgiosironi: unneeded import

profiles/logging.py
```diff
@@ -5,6 +5,9 @@
 
 from pythonjsonlogger import jsonlogger
 
+sql_engine_logger = logging.getLogger('sqlalchemy.engine')
+app_logger = logging.getLogger(__name__)
```
giorgiosironi: I think this is not equivalent to the original `logging.getLogger()`? That gets the root Logger object while this one gets the `profiles.logging` Logger, whose configuration won't cascade into other Logger objects

giorgiosironi: Unrelated: the global variables are not used so unclear whether they do need to be global, perhaps to be moved to another PR

docker-compose.dev.yml
GiancarloFusiello: OK, I'll update this.

profiles/config.py
GiancarloFusiello: updated

Dockerfile.dev
giorgiosironi: I meant `python:3.6.4-slim-jessie` which is an official one, also the eLife base image can work but it's not mandatory

Pipfile.lock
giorgiosironi: We'll take a decision on `pip` vs `pipenv` vs. anything else later, I think this is the best we can do with the current setup (ranting against `pipenv` I mean)

Dockerfile.dev
GiancarloFusiello: OK, I'll try the official python image first.

Pipfile.lock
GiancarloFusiello: For now, I can revert the changes to the Pipfile.lock and manually add the new entries. What do you think?

giorgiosironi: I don't trust Pipfile.lock to be manually editable so should be fine as it is

Dockerfile.dev
GiancarloFusiello: After testing, it seems that the full python is required to install dependencies.

docker-compose.dev.yml
GiancarloFusiello: I have added this change.

profiles/logging.py
GiancarloFusiello: I have reverted the code and, after resolving other configuration issues, this appears to be working as expected.

GiancarloFusiello: I have reverted this change.

Makefile
```diff
@@ -0,0 +1,22 @@
+help:
```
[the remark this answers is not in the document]
giorgiosironi: These ones have to be rewritten using `make` commands:
https://github.com/elifesciences/profiles#installation
https://github.com/elifesciences/profiles#running-the-tests
https://github.com/elifesciences/profiles#running-the-site
Remove:
https://github.com/elifesciences/profiles#local-virtual-environment-for-ide-usage

docker-compose.dev.yml
```diff
@@ -0,0 +1,38 @@
+version: '3'
```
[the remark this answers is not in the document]
GiancarloFusiello: done

Makefile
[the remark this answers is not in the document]
GiancarloFusiello: done

profiles/repositories.py
```diff
@@ -79,6 +80,8 @@ def remove(self, orcid: str) -> None:
             orcid_token = self.db.session.query(OrcidToken).filter_by(orcid=orcid).one()
             self.db.session.delete(orcid_token)
         except NoResultFound:
+            msg = 'Unalbe to remove ORCID token for ORCID {}. Token not found'.format(orcid)
```
giorgiosironi: typo in error message

profiles/repositories.py
```diff
@@ -68,8 +68,9 @@ def get(self, orcid: str) -> OrcidToken:
         try:
             return self.db.session.query(OrcidToken).filter_by(orcid=orcid).one()
         except NoResultFound as exception:
-            raise OrcidTokenNotFound('ORCID token for the ORCID {} not found'.format(orcid)) \
-                from exception
+            msg = 'ORCID token for ORCID {} not found'.format(orcid)
+            LOGGER.info(msg=msg)
```
giorgiosironi: consider logging some of these events as `error` rather than `info` if they are the result of exceptions (or are correlated with those)

profiles/repositories.py
```diff
@@ -102,16 +105,18 @@ def add(self, profile: Profile) -> Profile:
             self.db.session.rollback()
 
             message = exception.args[0]
+            log_msg = 'Unable to create profile with id {}. '.format(profile.id)
 
             if 'orcid' in message and profile.orcid:
-                LOGGER.info('Profile for ORCID %s appears to already exist', profile.orcid)
+                log_msg += 'Profile with ORCID {} already exists'.format(profile.orcid)
```
giorgiosironi: abbreviations like `msg` are usually unnecessary, at least the style of this file (existing variables) suggests to use `message` instead

giorgiosironi:
In general these should be two independent pull requests (development container and logging improvements), but _right now_ the easiest way forward is to keep them together here.

profiles/models.py
```diff
@@ -2,8 +2,8 @@
 from datetime import datetime
 from typing import Any, Iterable, List, Optional
 
-from iso3166 import Country
 import pendulum
+from iso3166 import Country
```
giorgiosironi: the linter (see build output) has opinions on import order and structure
```
profiles/models.py:6:1: I100 Import statements are in the wrong order. 'from iso3166 import Country' should be before 'import pendulum' and in a different group.
```
plus some other failing checks from static analysis

GiancarloFusiello:
@giorgiosironi I agree this should be two independent pull requests. I'll leave this for now but separate on future PRs.

profiles/models.py
GiancarloFusiello: I would disagree with the linter here. Is it ok if I update the linting rules?

profiles/repositories.py
GiancarloFusiello: The logger has an attribute called `msg` so I was following the same naming convention. I can change it if you prefer?

profiles/models.py
GiancarloFusiello: On second thought, i'm going to revert changes to imports for now and use a new format on newer projects.

profiles/repositories.py
de-code: Just wondering, whether logging exception messages adds much. Shouldn't it be logged anyway, where the exception is handled?

giorgiosironi: I thought so but couldn't find them when we needed, so erring on the side of caution

profiles/repositories.py
giorgiosironi: good if it follows an established convention

giorgiosironi:
Pylint has some complaints, easy to fix:
https://alfred.elifesciences.org/blue/organizations/jenkins/pull-requests-projects%2Fprofiles/detail/PR-103/7/pipeline

profiles/repositories.py
GiancarloFusiello: I also do not see these exceptions when looking through the logs. I suspect that they are more for generating HTTP responses.

de-code: Maybe the logging could be hooked in at a higher level to catch more errors?

giorgiosironi: https://rpm.newrelic.com/accounts/1451451/applications/63869969/traced_errors

PR #104, opened 2018-12-18
Add more logging to help with troubleshooting

GiancarloFusiello:


profiles/api/oauth2.py
```diff
@@ -98,21 +98,34 @@ def _check() -> Response:
...
         try:
             client = clients.find(request.form.get('client_id'))
         except KeyError as exception:
+            LOGGER.error('Invalid Client: %s not in list of known clients',
+                         request.form.get('client_id'))
             raise InvalidClient from exception
 
         redirect_uri = request.form.get('redirect_uri')
         if request.form.get('client_secret') != client.client_secret:
+            LOGGER.error('Invalid Client: client_secret not in request')
```
giorgiosironi: maybe this error messages should use the precise class name e.g. `InvalidClient` to make it clear is something you can catch and/or search for in the code rather than just a log message

profiles/api/webhook.py
```diff
@@ -19,9 +22,12 @@ def create_blueprint(profiles: Profiles, orcid_config: Dict[str, str],
 
     @blueprint.route('/orcid-webhook/<payload>', methods=['POST'])
     def _update(payload: str) -> Response:
+        LOGGER.info('POST request made to /orcid-webhook/%s', payload)
+
         try:
             orcid = uri_signer.loads(payload)
```
giorgiosironi: can we log anything about `orcid` here with `info` level? `payload` is unreadable probably

profiles/commands.py
```diff
@@ -87,5 +91,7 @@ def _convert_orcid_date(orcid_date: dict) -> Optional[Date]:
         year = int(orcid_date['year']['value'])
         month = int(orcid_date['month']['value']) if orcid_date.get('month') else None
         day = int(orcid_date['day']['value']) if orcid_date.get('day') else None
-
-        return Date(year, month, day)
+        try:
+            return Date(year, month, day)
+        except ValueError as exception:
+            LOGGER.error('%s: %s', exception, (year, month, day))
```
giorgiosironi: thus changes the behavior as unless `exception` is re-raised it will return `None`?

profiles/api/webhook.py
```diff
@@ -11,6 +12,8 @@
 from profiles.orcid import OrcidClient
 from profiles.repositories import OrcidTokens, Profiles
 
+LOGGER = logging.getLogger()
```
giorgiosironi: should use `getLogger(__name__)`? Same for the other calls introduced, but I'm not sure what the rest of the application is doing

profiles/commands.py
GiancarloFusiello: This is only used to set a keyword argument value: `starts=_convert_orcid_date(orcid_affiliation.get('start-date') or {}),`. As the caller expects a None value if the year is not set, I think raising an exception here shouldn't be breaking. What do you think?

profiles/api/webhook.py
GiancarloFusiello: I remember you said that I should use the root logger on another PR. I'll update these to use the module logger.

profiles/api/oauth2.py
GiancarloFusiello: ok, i'll convert for errors.

profiles/commands.py
giorgiosironi: If the caller can handle `None`, this is an actual bug fix probably as I agree in case of bad data coming in the best we can do is to log errors and skip those values (I remember someone seeing a 31st November or similar date). "breaking" is subtle as this was possibly raising an exception before (`ValueError`) and now won't be, it's a change that is probably for the best in this case.

(in general we tend to separate PRs by topic to keep them independent, improve logging vs. bugfix)

profiles/api/webhook.py
GiancarloFusiello: I added the `loads_unsafe` for logging purposes. This should translate the hashed payload into an ORCID without raising an exception.

profiles/commands.py
GiancarloFusiello: Agreed this should probably be in a separate PR. Would you like me to remove it and create a new PR?

giorgiosironi: Fine to keep it here for now, look out for similar cases into the future

PR #105, opened 2018-12-20
select orcid id from uri_signer tuple in log messages

GiancarloFusiello:


profiles/api/webhook.py
```diff
@@ -23,13 +23,13 @@ def create_blueprint(profiles: Profiles, orcid_config: Dict[str, str],
...
         LOGGER.info('POST request received from /orcid-webhook/%s',
-                    uri_signer.loads_unsafe(payload))
+                    uri_signer.loads_unsafe(payload)[-1])
 
         try:
             orcid = uri_signer.loads(payload)
         except BadSignature as exception:
             LOGGER.error('BadSignature: payload signature does not match: /orcid-webhook/%s',
-                         uri_signer.loads_unsafe(payload))
+                         uri_signer.loads_unsafe(payload)[-1])
```
giorgiosironi: `-1` or `1` (second or last)?

GiancarloFusiello: technically the same as two objects are returned. The only difference is that `-1` will always return the last object where `1` will only try to return the second and raise an exception if a second object is not returned. I think using `1` is more precise but could raise an exception and I wouldn't want the application to break due to logging. What do you think?

PR #106, opened 2018-12-20
allow authenticated requests to /profiles/<id> to return restricted data

GiancarloFusiello:


profiles/api/api.py
```diff
@@ -62,7 +63,10 @@ def _get(profile_id: str) -> Response:
         except ProfileNotFound as exception:
             raise NotFound(str(exception)) from exception
 
-        response = make_response(json.dumps(profile, default=normalize))
+        if request.headers.get('x-consumer-groups') == 'view-restricted-profiles':
```
giorgiosironi: this header may have the form `view-restricted-profiles, view-something-else-unrelated, ...` so it needs to be parsed

profiles/api/api.py
```diff
@@ -62,7 +63,10 @@ def _get(profile_id: str) -> Response:
...
+            response = make_response(json.dumps(profile, default=normalize_restricted))
+        else:
+            response = make_response(json.dumps(profile, default=normalize))
```
giorgiosironi: I'm not sure if the fields involved are appearing in `/profiles`, it may need to be checked there too

In general I've seen in other services that the header is handled for all APIs:
https://github.com/elifesciences/digests/blob/4b418049799a999193e5ed4eb80b7587a123b0d8/app/core/middleware.py#L22

profiles/serializer/normalizer.py
```diff
@@ -12,6 +12,11 @@ def normalize(value: Any) -> Any:
     return value
 
 
+@singledispatch
+def normalize_restricted(value: Any) -> Any:
```
giorgiosironi: is the additional layer of indirection a type checker/linter requirement? (allows `Any` rather than just `Profile`?)

profiles/api/api.py
giorgiosironi: Not sure if case insensitivity (`X-Consumer-Groups` could be passed in) is handled by Flask for headers? It's part of the standard so I would expect so

test/test_api.py
```diff
@@ -314,3 +315,54 @@ def test_it_does_not_return_restricted_affiliations(test_client: FlaskClient, ye
...
+    assert validate_json(data, schema_name='profile.v1') is True
+    assert len(data['affiliations']) == 1
+    assert len(data['emailAddresses']) == 1
+    assert [e['value'] for e in data['emailAddresses']] == ['1@example.com']
+
+
+def test_it_does_not_return_restricted_data_when_authenticated(
+        test_client: FlaskClient, yesterday: Date, commit: Callable[[], None]) -> None:
+
+    address = Address(country=countries.get('gb'), city='City', region='Region')
```
giorgiosironi: the setup of these methods (until `commit()`) looks duplicated, could be extracted into a `conftest` fixture for readability of test and for its reuse

profiles/api/api.py
GiancarloFusiello: very good points. I'll update.

GiancarloFusiello: Calling `get` on the `headers` object is case insensitive so i'll update the logic to handle the value which is returned in it's original case.

profiles/api/api.py
GiancarloFusiello: `/profiles` returns the results of calling `Profiles.list`. This in turn does not have a parameter to specify restricted data. This may raise the question of, if an ORCID user has their profile set with `trusted` permissions for their `name` and `id`, do we expected to see this in the Profiles database?

test/test_api.py
GiancarloFusiello: We could do that but my preference is to view this information in the test so I can see what is the state of the database. For the two tests in this PR the data is almost the same (one is set to restricted and the other not) but other tests may have different setup fixtures. Another approach is the use factory objects that create db entires with minimal syntax but is fully customisable. However, given the number of tests already written, this may be overkill and more suited to a younger project.

profiles/serializer/normalizer.py
GiancarloFusiello: I tried to follow the current pattern in place. `@singledispatch` is a way of using Generics (overloading) in python. Meaning, a call to `normalize` will first check the type of the first argument and, if another registered function exists that expects that type, then that implementation will be used.

profiles/api/api.py
giorgiosironi: I remember [handling of private names](https://github.com/elifesciences/profiles/pull/31) being introduced but not of `trusted` ones. If we go for the same behavior we would reject the user? Probably needs a check on existing profiles but I think it's difficult to correlate (if we don't store that information we will need to call ORCID with all the various tokens).
I also found the [original implementation](https://github.com/elifesciences/profiles/pull/58) of the restricted access if that helps with context.

GiancarloFusiello: I've spoken with @thewilkybarkid and we think that it is only needed here for the time being. If there are more use cases then this can be extracted into middleware.

PR #109, opened 2019-01-08
Remove null values from profiles response

GiancarloFusiello:
Currently, `/profiles` will return `null` values for optional fields that do not have a value which invalidates the schema.

- remove null values from `/profiles` response

Related to issue [#3365](https://github.com/elifesciences/issues/issues/3365)

### DIFF
--- a/test/utilities/test_contains_none_values.py
+++ b/test/utilities/test_contains_none_values.py
@@ -30,7 +30,7 @@ LIST_WITH_NESTED_LISTS_WITH_NONE_VALUE_AND_FOLLOWING_VALUES = [
     4]
 
 
-@pytest.mark.parametrize('input, expected_result', [
+@pytest.mark.parametrize('test_data, expected_result', [
     (EMPTY_DICT, False),
     (DICT_WITH_NONE_VALUE, True),
     (DICT_WITH_NESTED_DICTS_WITHOUT_NONE_VALUE, False),
@@ -41,5 +41,5 @@ LIST_WITH_NESTED_LISTS_WITH_NONE_VALUE_AND_FOLLOWING_VALUES = [
     (LIST_WITHOUT_NONE_VALUE, False),
     (LIST_WITH_NESTED_LISTS_WITH_NONE_VALUE_AND_FOLLOWING_VALUES, True)
 ])
-def test_contains_none_values(input, expected_result) -> None:
-    assert contains_none_values(input) is expected_result
+def test_contains_none_values(test_data, expected_result) -> None:
+    assert contains_none_values(test_data) is expected_result


profiles/serializer/normalizer.py
```diff
@@ -44,21 +43,29 @@ def normalize_restricted_profile(profile: Profile) -> dict:
 
 @normalize.register(Affiliation)
 def normalize_affiliation(affiliation: Affiliation) -> dict:
+    if affiliation.department is not None:
+        name = [affiliation.department, affiliation.organisation]
```
giorgiosironi: I wonder if you could build `name` and `components` with the mandatory fields first, and then add the optional values in the right place (0 index and `area` field, respectively). It removes duplication between the `if` branches but it's more complex.

profiles/serializer/normalizer.py
```diff
@@ -44,21 +43,29 @@ def normalize_restricted_profile(profile: Profile) -> dict:
...
+        name = [affiliation.organisation]
+
+    if affiliation.address.region is not None:
+        address = [affiliation.address.city, affiliation.address.region,
+                   affiliation.address.country.name]
+        components = {"locality": [affiliation.address.city],
+                      "area": [affiliation.address.region],
+                      "country": affiliation.address.country.name}
+    else:
+        address = [affiliation.address.city, affiliation.address.country.name]
```
giorgiosironi: `address` could be derived from `components`, since it's conceptually a human-readable version of that?

test/test_api.py
```diff
@@ -367,3 +367,35 @@ def test_it_does_not_return_restricted_data_when_authenticated(
...
+
+    headers = Headers()
+    headers.set('X-Consumer-Groups', 'View-restricted-profiles, Something else')
+    response = test_client.get('/profiles/a1b2c3d4', headers=headers)
+    data = json.loads(response.data.decode('UTF-8'))
+
+    assert response.status_code == 200, response.status_code
+    assert response.headers.get('Content-Type') == 'application/vnd.elife.profile+json;version=1'
+    assert validate_json(data, schema_name='profile.v1') is True
+    assert dict_contains_none_values(data) is False
```
giorgiosironi: good test case

profiles/utilities.py
```diff
@@ -80,6 +80,21 @@ def remove_none_values(items: dict) -> dict:
     return dict(filter(lambda item: item[1] is not None, items.items()))
 
 
+def dict_contains_none_values(data: dict) -> bool:
+    visited, stack = [], list(data.values())
+    while stack:
+        vertex = stack.pop()
+        if vertex is None:
+            return True
+        if vertex not in visited:
```
giorgiosironi: I think `visited` caters for a dictionary having a reference to itself (a graph), but this should be a tree of values so it wouldn't be necessary?
In general a recursive implementation would be much shorter if it's equivalent:
```
def contains_none_values(data):
    for element in data:
        if element is None:
            return True
       if isinstance(element , dict):
            if contains_none_values(element.values()):
                return True
       if isinstance(element , list):
            if contains_none_values(element):
                return True
return False
```

profiles/serializer/normalizer.py
GiancarloFusiello: good idea. updated.

profiles/serializer/normalizer.py
GiancarloFusiello: updated.

test/test_api.py
GiancarloFusiello: thanks

profiles/utilities.py
GiancarloFusiello: Usually I try to iterate rather than make recursive calls due to pythons recursion limit, however, I think you're right here as the use case for this is for smaller objects rather than large graphs. I have updated the code to use accordingly.

test/utilities/test_contains_none_values.py
```diff
@@ -0,0 +1,34 @@
+from profiles.utilities import contains_none_values
+
+
+def test_contains_none_values() -> None:
```
de-code: personally I'd probably split into separate tests

de-code:
I had a similar issue in PeerScout a while back, where the generated JSON was even generating `None` instead of `null`. Similar to your `contains_none_values` I was then removing `None` values before it became the API's result. But you'll probably have fewer case where that could happen.

test/utilities/test_contains_none_values.py
thewilkybarkid: Should be parametrised, so we can see all the failures rather than just 1.

profiles/serializer/normalizer.py
```diff
@@ -44,21 +43,23 @@ def normalize_restricted_profile(profile: Profile) -> dict:
...
-                    affiliation.address.city,
-                    affiliation.address.region,
-                    affiliation.address.country.name,
-                ],
-                "components": {
-                    "locality": [affiliation.address.city],
-                    "area": [affiliation.address.region],
-                    "country": affiliation.address.country.name
-                }
+                "formatted": [v[0] if isinstance(v, list) else v for v in components.values()],
```
giorgiosironi: this looks general, cool

test/utilities/test_contains_none_values.py
giorgiosironi: https://docs.pytest.org/en/latest/reference.html#pytest-mark-parametrize-ref ?

de-code: yes, parameterise is good too. Separate methods would have the advantage of better expressing what they are testing. e.g. what makes `dict_3` and `dict_4` different..

GiancarloFusiello: How does adding `pytest.mark.parametrize` improve this test?

GiancarloFusiello: @de-code `dict_3` does not contain a `None` value and, when passed to the function, should return `False`. `dict_4` does contain a `None` value and, when passed to the function, should return `True`. Do we need separate tests to explain this?

de-code: @GiancarloFusiello I might be able to infer it from looking at it a bit longer but it's not so clear when the intention is not expressed. There could also be a _bug_ in the test data. Could go further, why do we need `dict_4` and `dict_5` when we have `dict_2`. I could guess what the reason was but it's getting less clear with `dict_5`. If I had to modify the behaviour in the future it wouldn't be as easy to tell whether the test is still valid. To me these are all separate test cases.

Another way to express the intention more would be to name the variables more distinctly (or use constants). e.g. `DICT_WITH_NESTED_NONE_VALUE` (or something like that)

Currently the test input data (`dict_1` etc) and expected output (`True`/`False`) seem to be a bit far apart. They would naturally sit close if they were in a separate test method but could be moved together if not.

`pytest.mark.parametrize` would also move them together and run each test individually (you would parameterise a list of _input data_ and _expected output_ pairs). It could then run all of the tests even if one was failing. I find parameterise may be more clear where the expected output is the same or similar. e.g. input values with different character sets. Or maybe where the test method is a bit more complex to avoid repetition.

Just my thoughts (maybe I just need more sleep).

GiancarloFusiello: @de-code I understand now. The variable names used in the test case do not describe the value. You're right, I'll update these.

Admittedly not having used pytest much in the past, I didn't fully understand the benefits of using `pytest.mark.parametrize` by reading the documentation but, after having found some examples, I can see how this can be useful.

I'll update the code and push these changes.

test/utilities/test_contains_none_values.py
```diff
@@ -1,34 +1,45 @@
...
+    (DICT_WITH_NESTED_DICTS_WITHOUT_NONE_VALUE, False),
+    (DICT_WITH_NESTED_DICTS_WITH_NONE_VALUE, True),
+    (DICT_WITH_NESTED_DICTS_WITH_NONE_VALUE_AND_FOLLOWING_VALUES, True),
+    (DICT_WITH_NESTED_DICT_AND_LIST_WITH_NONE_VALUE, True),
+    (LIST_WITH_NONE_VALUE, True),
+    (LIST_WITHOUT_NONE_VALUE, False),
+    (LIST_WITH_NESTED_LISTS_WITH_NONE_VALUE_AND_FOLLOWING_VALUES, True)
+])
+def test_contains_none_values(input, expected_result) -> None:
+    assert contains_none_values(input) is expected_result
```
de-code: minor comment: input is a special keyword. might be good to rename

GiancarloFusiello: Good point.

PR #110, opened 2019-01-08
Update readme with basic project description and end point information

GiancarloFusiello:


README.md
```diff
@@ -3,6 +3,26 @@ eLife Profiles
...
+When a user account is created or updated on ORCID, the new account information is sent to 
+the Profiles service via a webhook and a new eLife profile is created, ready to be
+used by other services.
+
+End points
+----------
+- `/ping` returns the message `pong`
+- `/profiles` returns a list of profiles *(this is paginated by default, 
+sending 20 results per request. Only basic info sent. 
+See [Here](https://api.elifesciences.org) for more information)*
```
thewilkybarkid: I would prefer just linking to the API docs instead of including details. In the event we change the default this is another place to remember to change...

giorgiosironi: For pagination the API documentation is a good source of truth, whereas this is a good place for documenting the difference basic and private information; this part is not covered by the API documentation as it states there is a security scheme based on `Authorization` but not what it's used for.

thewilkybarkid: Would be better to have that in the API docs. Not sure if RAML and/or the UI is too limiting.

giorgiosironi: Mostly yes, meant it's not there at the moment, and `profiles` can't know anything about `Authorization` as it's dealt with in the `api-gateway` and hidden from it. If there is any ORCID-specific information however like a mapping between private/trusted/public and what happens with them, it's part of this implementation.

GiancarloFusiello: I've just updated this. Any better? As it's just documentation, feel free to push your own changes/suggestions.

README.md
```diff
@@ -3,6 +3,23 @@ eLife Profiles
 
 [![Build Status](https://ci--alfred.elifesciences.org/buildStatus/icon?job=test-profiles)](https://ci--alfred.elifesciences.org/job/test-profiles/) [![Coverage Status](https://coveralls.io/repos/github/elifesciences/profiles/badge.svg?branch=develop)](https://coveralls.io/github/elifesciences/profiles?branch=develop)
 
+Profiles is a service used by eLife Sciences Publications, Ltd. to create and 
+store eLife user profiles. Currently, eLife uses [ORCID](https://orcid.org/) for 
+authentication.
+
+When a user account is created or updated on ORCID, the new account information is sent to
```
giorgiosironi: Incorrect: we setup one webhook per user when we create the account, so the flow is:
- when a user authenticates through profiles using ORCID OAuth, a user account is created; or updated if existing
- when a user account is updated on ORCID, the webhook notifies this service to update the existing user account

The main point being we create a user only through voluntary authentication, with a very small subset of the ORCID population that happens to use elifesciences.org or other services.

Final point to note: it is possible to import users manually inside this service without going through the ORCID authentication, so profiles are not necessarily tied to an ORCID account. They will never get updated in that case (it's for imports of historical data).

GiancarloFusiello:
I'm going to close this PR and suggest someone with access to/knowledge of the original spec for this service adds a description of the function of this service.

lsh-0:
re-opening this as the current state of the documentation for this project is woeful.

PR #111, opened 2019-01-14
security fix, bumps pyyaml to 4.2b

lsh-0:
* pins version of sqlalchemy to stop it being bumped a major version

vulnerable usage here:
https://github.com/elifesciences/profiles/blob/develop/manage.py#L19

cc @GiancarloFusiello @giorgiosironi


lsh-0:
@GiancarloFusiello , I'm not sure what your stance is on pipenv, but I find it a total pain in the arse. profiles is one of a few (perhaps the only) that uses it.

giorgiosironi:
The error in the test suite
```
OverflowError: date value out of range
```
initially looked like tests depending on the current time, but it's repeated across several tests so it may be genuine.

GiancarloFusiello:
> @GiancarloFusiello , I'm not sure what your stance is on `pipenv`, but I find it a total pain in the arse. `profiles` is one of a few (perhaps the only) that uses it.

In theory `pipenv` sounds great but in practice, not so much. I have been looking at an alternative called `poetry` which, although not perfect, is pretty good. Perhaps we can go over this in a google hangout sometime. If you like it we can look at adopting it in this project and future projects. Either way, we should probably move this conversation offline/in a different PR.

Pipfile.lock
```diff
@@ -536,17 +530,16 @@
         },
         "ipdb": {
             "hashes": [
-                "sha256:7081c65ed7bfe7737f83fa4213ca8afd9617b42ff6b3f1daf9a3419839a2a00a"
+                "sha256:aa329b7fed373b78778fbffc53245761a94e0648086b7924e92048a0467e4763"
             ],
             "index": "pypi",
-            "version": "==0.11"
+            "version": "==0.8.3"
```
GiancarloFusiello: This is why I don't like `pipenv`. Why has an unrelated package been updated?!

GiancarloFusiello:
@lsh-0 I hope you don't mind, I saw that the tests were failing. This was due to the hypothesis package being updated which must change the behaviour in some way.

My commit simply reverts most of the packages that shouldn't have been updated by pipenv but still updates PyYAML.

lsh-0:
yes, I figured it was the Hypothesis dep that was causing the issue. I didn't want to do piecemeal reversion of the unrelated dependencies because there is a global sha in the `.lock` file that breaks when you do that (and I couldn't figure out how to target an individual dep).

I'm fine with the changes and will merge.

Pipfile.lock
lsh-0: well, we know why, we just don't like it ;) I reckon this must be how Assembly programmers felt when C came along ...

giorgiosironi:
Just ponting out for anyone reading that we are talking about [Hypothesis the property-based testing framework](https://hypothesis.works/) and not [Hypothesis the annotation service](https://web.hypothes.is/).

lsh-0:
oh - I just assumed it was some annotation library thing. Thanks for clearing that up

PR #151, opened 2019-03-14
fix db getting into broken state

GiancarloFusiello:
Relates to https://github.com/elifesciences/issues/issues/4633

Fix for db getting into broken transaction state.

Seems to be an issue with either Flask-SQLAlchemy or SQLAlchemy. Either way, no exception is raised by either framework. The problem becomes apparent when trying to access the database after the database is in a broken state.

I have tried updating all dependencies to the latest version but the problem still persists.

My solution is to add a check to see if an email in an incoming ORCID record tries to update a profile with an email associated with another profile. If the email already exists the error is logged and the email is removed from the list of emails to process/update.

### DIFF
--- a/profiles/commands.py
+++ b/profiles/commands.py
@@ -4,8 +4,10 @@ from typing import List, Optional
 from iso3166 import countries
 import jmespath
 
-from profiles.models import Address, Affiliation, Date, Name, Profile
+from profiles.exceptions import ProfileNotFound
+from profiles.models import Address, Affiliation, Date, Name, Profile, db
 from profiles.orcid import VISIBILITY_PUBLIC
+from profiles.repositories import SQLAlchemyProfiles
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


test/test_webhook.py
```diff
@@ -35,6 +36,59 @@ def test_it_updates_and_returns_204_if_a_profile_is_found(test_client: FlaskClie
     assert profile.name.preferred == 'Given Names Family Name'
 
 
+def test_it_does_not_cause_db_to_get_into_broken_state(test_client: FlaskClient,
```
giorgiosironi: to the best of my knowledge is the db _connection_ getting into a state where no more transactions can be run?
Would also be helpful to paste into the docstring of this test what we know about the error, since it makes it searchable in case it happens again

thewilkybarkid: :+1: Given this is a bit of a(n unfortunately necessary) hack, it'd be useful to have a bit more context here.

GiancarloFusiello: Renamed test and added link to github issue.

profiles/commands.py
```diff
@@ -70,18 +72,34 @@ def _update_affiliations_from_orcid_record(profile: Profile, orcid_record: dict)
...
+            email_profile = profiles.get_by_email_address(email)
+        except ProfileNotFound:
+            continue
+
+        if email_profile.id != profile.id:
+            message = ('Profile %s is trying to add email address %s but this '
+                       'email is associated with profile %s which violates '
+                       'unique constraint for email addresses' %
+                       (profile.id, email, email_profile.id))
+            LOGGER.error(message)
```
giorgiosironi: Just being aware of the fact that this error will be logged every time an update for `profile.id` is delivered? But since we don't fail the request, it's not very frequent.

test/test_webhook.py
```diff
@@ -35,6 +36,65 @@ def test_it_updates_and_returns_204_if_a_profile_is_found(test_client: FlaskClie
...
+                                   {'email': '1@example.com',
+                                    'primary': True,
+                                    'verified': True,
+                                    'visibility': 'PUBLIC'}
+                               ]
+                           }
+                       }
+                   })
+
+        response = test_client.post('/orcid-webhook/{}'.format(webhook_payload))
```
giorgiosironi: `webhook_payload` is opaque here (the test code doesn't tell us which ORCID it corresponds to); maybe it can be forced to be about `0000-0002-1825-0097`

profiles/commands.py
GiancarloFusiello: It's only logged when the condition is True on Line 85.

test/test_webhook.py
```diff
@@ -35,6 +36,65 @@ def test_it_updates_and_returns_204_if_a_profile_is_found(test_client: FlaskClie
...
+
+    response = test_client.get('/profiles/b2c3d4e5')
+    assert response.status_code == 200
+
+    data = json.loads(response.data)
+    assert data['id'] == 'b2c3d4e5'
+    assert data['orcid'] == '0000-0002-1825-0097'
+    assert data['emailAddresses'] == []
+    assert data['name']['preferred'] == 'New Name'
+# pylint: enable=invalid-name
```
giorgiosironi: what name it was this time? :sweat_smile:

test/test_webhook.py
GiancarloFusiello: It's a serialised version of `0000-0002-1825-0097` https://github.com/elifesciences/profiles/blob/develop/test/conftest.py#L204. This pattern is used throughout the test suite.

test/test_webhook.py
GiancarloFusiello: After changing the test name apparently the name was too long.

PR #161, opened 2019-04-17
update outdated packages

GiancarloFusiello:
PR was to update Jinja2 to version 2.10.1 for https://github.com/elifesciences/issues/issues/4874 but appears to be difficult with pipenv as it will update other packages also. However, tests run without issue for dev environment.

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "a003acc0a116bdf1815518ced2dd4f40f2e894caf96362a68e3c4aa73b2315d9"
+            "sha256": "9e1f9c1988006138a72c0d869edfe772aa0c6874127d87b94b19bea17ab29d5a"
         },
         "pipfile-spec": 6,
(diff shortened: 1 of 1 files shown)


giorgiosironi:
Doesn't touch `SQLAlchemy` which is listed at https://github.com/elifesciences/profiles/network/alerts; not sure if it should be in another PR or in this one.

GiancarloFusiello:
> Doesn't touch `SQLAlchemy` which is listed at https://github.com/elifesciences/profiles/network/alerts; not sure if it should be in another PR or in this one.

I merged before I saw this comment so it will be in another PR now. Again, I get a 404 for https://github.com/elifesciences/profiles/network/alerts suggesting I do not have access to that page. Is there anything else that needs updating?

giorgiosironi:
> I get a 404

@elifesciences/tech was added with `Write` permissions, I added you as `Admin`; probably for security reasons the alerts are limited in visibility. Only`Upgrade SQLAlchemy to version 1.3.0 or later` is listed now.

PR #164, opened 2019-04-23
update urllib3 >= 1.24.2 as per github security alert CVE-2019-11324

GiancarloFusiello:


Pipfile
```diff
@@ -25,7 +25,7 @@ PyYAML = "==4.2b4"
 requests = "~=2.18"
 retrying = "~=1.3"
 SQLAlchemy = "~=1.3"
-urllib3 = "~=1.22"
+urllib3 = "~=1.24"
```
giorgiosironi: I would have thought `>=1.24.2` but that also needs a constraint like `<2` to avoid a new major version. I think this one is fine.

Probably would have upgraded to `1.24.2` also if it was remaining `~=1.22` if I understand correctly, interpreting that as `>=1.22,<2`?

